PR #3, opened 2021-07-27
Brush up design documents

BarbarossaTM:
* Fix some typos
 * All acronys in capital letters

### DIFF
--- a/docs/design/01-address-scope.md
+++ b/docs/design/01-address-scope.md
@@ -30,7 +30,7 @@ Address Scopes in CCloud define the relationship of a subnet (via the indirectio
 ### Driver Config
 ```
 [global]
-base-asn = 65130 # Example QA-DE-1
+base-asn = 65130 # Example qa-de-1
 [vrf-mapping]
 cc-cloud01 = 3b189848-58bb-4499-abc2-8df170a6a8ae, 4143da3e-d2a7-4077-ba80-215ecfd016d7
 
@@ -63,7 +63,7 @@ Core originated subnets
   * eOS
   * NX-OS 
 
-Example based on Leafs in QA-DE-1a, assuming QA-DE-1 has AZ's a,b,d
+Example based on Leafs in qa-de-1a, assuming qa-de-1 has AZ's a,b,d
 
 ### Generic Pod Local Leaf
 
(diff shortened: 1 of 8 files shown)


docs/design/01-address-scope.md
```diff
@@ -63,7 +63,7 @@ Core originated subnets
   * eOS
   * NX-OS 
 
-Example based on Leafs in QA-DE-1a, assuming qa-de-1 has AZ's a,b,d
+Example based on Leafs in QA-DE-1a, assuming QA-DE-1 has AZ's a,b,d
```
sebageek, line 66: qa-de-1 should be consequently written lowercase, at least that's how I like it. Let's make it consistent in this direction.

docs/design/03-ports-and-port-bindings.md
```diff
@@ -47,8 +49,8 @@ Config on NXOS leaf:
 ```
 
 
-## Baremetal on Fabric
-Baremetal servers are in most cases directly on dedicated Arista switches which will become part of our fabric.
+## Bare metal on Fabric
```
sebageek, line 52: Wouldn't the `m` in `metal` also be uppercase in a title?

docs/design/10-management-networks.md
```diff
@@ -1,12 +1,12 @@
...
+Originally these networks were seen as local to a leaf pair, but as we also want to be able to put servers from another
+leaf pair into these networks (e.g. in case of maintenance or hardware failure), we cannot treat these networks as
+local to a leaf pair.
 
 Between members of this network we need to have VLAN consistency (as we do when "lending a host" to another bb).
 This means each port in a management network will get it as the same VLAN (i.e. mgmt --> 100). It will still need
 to be represented with its own VNI inside the fabric.
 
-Management networks are l3, so they also need to have a gateway.
+Management networks are L3, so they also need to have a gateway.
```
sebageek, line 12: Why uppercase `L` here? would we capitalize Layer 3?

sebageek (changes requested):

docs/design/01-address-scope.md
BarbarossaTM: Will do

docs/design/03-ports-and-port-bindings.md
BarbarossaTM: Will do

docs/design/10-management-networks.md
BarbarossaTM: Because it's an acronym which we wrote in upper case in most places and to make it more readable and upper case L is easier distinguished from 1 and i as lower case :-)

PR #9, opened 2021-09-23
Agent switch connection

sebageek:


networking_ccloud/common/constants.py
```diff
@@ -12,8 +12,9 @@
 #    License for the specific language governing permissions and limitations
 #    under the License.
 
-# FIXME: cisco should be nxos, as we go by "os/api type"
-VENDORS = ["arista", "cisco"]
+VENDOR_ARISTA = "arista"
+VENDOR_NXOS = "nxos"
+VENDORS = [VENDOR_ARISTA, VENDOR_NXOS]
```
chrysh, line 17: Why not a dictionary?

sebageek: I use `VENDOR_ARISTA` and `VENDOR_NXOS` in the agents as constants in different places. I need a list of all vendors as well to iterate over / do valid vendor checks. How would this look like with a dictionary?

PR #11, opened 2021-09-27
Start rpc on external request, not on initialize()

sebageek:
Neutron will call this driver's get_workers(), on which then
start_rpc_listeners() will be called. We need to do it this way to allow
this driver to work with a split-api-rpc setup.


networking_ccloud/ml2/mech_driver.py
```diff
@@ -51,12 +50,27 @@ def _plugin(self):
             self._plugin_property = directory.get_plugin()
         return self._plugin_property
 
-    def start_driver_rpc(self):
+    def start_rpc_listeners(self):
+        """Start the RPC listeners.
+
+        Most plugins start RPC listeners implicitly on initialization.  In
+        order to support multiple process RPC, the plugin needs to expose
```
joker-at-work, line 57: multiple process RPC? I like the wording in the commit-message more. This one, I don't understand.

networking_ccloud/ml2/mech_driver.py
```diff
@@ -51,12 +50,27 @@ def _plugin(self):
...
 
         return self.conn.consume_in_threads()
 
+    def get_workers(self):
+        """Get any NeutronWorker instances that should have their own process
+
+        Any driver that needs to run processes separate from the API or RPC
+        workers, can return a sequence of NeutronWorker instances.
+        """
+        return [service.RpcWorker([self], worker_process_count=0)]
```
joker-at-work, line 72: The `worker_process_count=0` is a little confusing for someone without knowledge of the topic.

networking_ccloud/ml2/mech_driver.py
sebageek: For the MechanismDriver I've been copying the original docstring into my file whenever I implement one of the methods, so this is taken from here: https://github.com/openstack/neutron-lib/blob/master/neutron_lib/plugins/ml2/api.py#L398-L404

You think a custom™ message would make more sense?

joker-at-work: hm ... then I don't understand the original and would have to take that confusion to upstream, I guess. No sure if a custom message is necessary, I just don't understand this one :D

networking_ccloud/ml2/mech_driver.py
sebageek: Apparently this is the upstream way, e.g. https://github.com/sapcc/neutron/blob/ce33533c8a7e50c0b13546b5ca6836e3e14a87c3/neutron/services/l3_router/l3_router_plugin.py#L134

joker-at-work: Extremely confusing, but thank you for digging that out. :+1:

networking_ccloud/ml2/mech_driver.py
sebageek: Maybe plugins can start multiple RPCWorker for different processes and that's all they're trying to tell you? But I won't haggle with upstream for a docstring myself, still trying to get the segment race condition fixed ;)

PR #12, opened 2021-10-13
Port bindings

sebageek:
Let's try to get a working port binding.

### DIFF
--- a/networking_ccloud/common/exceptions.py
+++ b/networking_ccloud/common/exceptions.py
@@ -1,0 +1,20 @@
+# Copyright 2021 SAP SE
+#
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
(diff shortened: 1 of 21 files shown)


networking_ccloud/common/exceptions.py
```diff
@@ -0,0 +1,20 @@
...
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#    under the License.
+
+from neutron_lib import exceptions as n_exc
+
+
+class MultipleBindingHostsInBindingProfile(n_exc.NeutronException):
+    """An exception indicating multiple hosts found in binding profile"""
+    message = "Port %(port_id)s has multiple hosts in binding profile, refusing to work with it! Hosts were: %(hosts)s"
```
joker-at-work, line 20: Please change the `, refusing` to `. Refusing` or ` - refusing` or `; refusing`. Otherwise, the hosts would not like to work with the ports. My understanding is, that the driver doesn't like to work with the port.

PR #13, opened 2021-12-13
Fix PortBindingLevel join condition for multiple portbindings

sebageek:
Could use an additional test, but just to get it on the way.


sebageek:
Now it has an additional test!

PR #21, opened 2022-02-17
Initial documentation on netbox input for config generation

martink2:
Initial documentation on netbox input for config generation


doc/source/configuration/config-input.rst
```diff
@@ -0,0 +1,144 @@
...
+::
+    CC-NET-EVPN-DHCP-RELAY:10.10.10.10
+    CC-NET-EVPN-DHCP-RELAY:10.11.11.11
+
+L2 Networks VLAN to VNI mapping
+#################################
+Netbox does not yet support a moddel for overlay network VNI's, the follwoing conventions are used
+
+* **Infra Regional**: VLAN X uses VNI X (VLAN 100 -> VNI 100)
+* **Infra AZ-wide**: VLAN X uses VNI [AZ-Prefix]X (VLAN 800, AZ=a -> 10800, VLAN 800, AZ=b -> 20800)
```
sebageek, line 101:
```
from oslo_config import cfg
cfg.CONF.ml2_type_vxlan
<oslo_config.cfg.ConfigOpts.GroupAttr object at 0x7f905b6535f8>
cfg.CONF.ml2_type_vxlan.vni_ranges
['10000:20000']
```
10000-20000 is the default vni range of OpenStack in our environment. If you want to use this exact scheme then we need to renumber all existing toplevel OpenStack network segments in our env. We can do that, but we'd have to put this on our agenda. Would that be the plan then or were you not aware of this?

doc/source/configuration/config-input.rst
```diff
@@ -0,0 +1,144 @@
...
+* **Infra Pod-wide**: VLAN X re-used in many pods as local vlan 100 -> **TBD**
+* **Tenant**: CCloud platform driver should use range 1.000.000 - 1.100.000
```
sebageek, line 103: As said, then we probably need to do a renumbering. Is that the plan?

doc/source/configuration/config-input.rst
```diff
@@ -0,0 +1,144 @@
...
+
+
+*******************
+Ports 
+*******************
+the driver is responsible for front ports on pod equipment, some port types require 
+certain infra VLAN's to be provisioned as well as ports beeing assembled into port-channels
```
sebageek, line 110: `VLAN's` -> `VLANs`, `beeing` -> `being`.

doc/source/configuration/config-input.rst
```diff
@@ -0,0 +1,144 @@
...
+based on current port function
+
+Port infra VLANs
+###################
+Infra VLAN's required on ports are recorded on the netbox port they are reuqired on,
+for port-channels the reuqired vlans do only need to be provided on the LAG interface,
```
sebageek, line 116: `reuqired` -> `required`

doc/source/configuration/config-input.rst
```diff
@@ -0,0 +1,144 @@
...
+VLAN's defined on member interfaces will be ignored for port-channel members:
```
sebageek, line 117: Well, they will only be ignored as long as the port-channel is assembled. When the PC breaks up (e.g. on DHCP/PXE install or in the baremetal usecase where user breaks their config) these VLANs probably need to be configured on the child interfaces as well.

doc/source/configuration/config-input.rst
```diff
@@ -0,0 +1,144 @@
...
+::
+    static: port-channel1 - port-channel199
+    dynamic: port-channel200 - port-channel299
+
+Port-channels can either have ports only on one device or be spanned across two
+devices (MLAG/vPC) the following convention will be used to distinguish the two 
+variants:
+::
+    port-channel1 defined on device 1110a only: a regular port-channel will be configured
+    port-channel1 defined on device 1110a AND 1110b: a MLAG/vPC will be configured
```
sebageek, line 143: The implication of this is that we need to track in the driver config if something is supposed to be an mlag or not, right? Meaning I have to take this as an action point for later?

doc/source/configuration/config-input.rst
```diff
@@ -0,0 +1,144 @@
...
+Leaf Pair
+************
+
+All EVPN fabric members will be of the following Netbox roles:
+
+* **EVPN Spine**: Used for all spine switches, not configured by driver
+* **EVPN Border Gateway**: Used for all gateways interconnecting CC AZ's in a multi AZ fabric
+* **EVPN Leaf**: Used for all leaf switches in the EVPN fabric
+
+Further more specific leaf functions will be identified by adding tags to the leafs.
```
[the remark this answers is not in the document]
martink2: role=netbox role (leaf|spine|bgw)
function = Border Leaf| Transit leaf| etc.

doc/source/configuration/config-input.rst
```diff
@@ -0,0 +1,144 @@
...
+Netbox Query:
+::
+    Role: EVPN Leaf
+    Name: .*(\d\d\d\d)([a-b]).* #\1=leaf_id \2=leaf_a_b
+    Tag: CC-APOD|CC-BPOD|CC-NETPOD|CC-STPOD|CC-VPOD
+
+CND EVPN leaf types:
+####################
+The driver needs to identify certain non-pod leaf pairs to 
+push tenant configuration:
```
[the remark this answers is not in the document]
martink2: depends , i suppose so , this was meant for the case border leaf where bgp summery config must go on for subnet pools but nothing else driver controlled

doc/source/configuration/config-input.rst
```diff
@@ -0,0 +1,144 @@
...
+
+If for a device a different (or reduced) range is in effect it must be expressed in the Netbox device config context for that device
+.. code-block::
+    {
+    "cc": {
+        "net": {
+            "evpn": {
+                "tenant-vlan-range": [
+                    "2000",
+                    "2100-3000"
```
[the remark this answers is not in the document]
martink2: i would assume a list of either single values or ranges and including

doc/source/configuration/config-input.rst
martink2: yep

doc/source/configuration/config-input.rst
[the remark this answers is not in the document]
sebageek: Ah, so they have some overlap, but not entirely? We might have a more thorough discussion about that then at some point, I might have misnamed some things to `role` in code. But we'll see ;)

PR #35, opened 2022-05-12
Overhaul config generation, add transits and bgw

swagner-de:


sebageek (approved):
Didn't look 100% through the config gen, but big picture looks great. :)

PR #36, opened 2022-05-16
Use shorthands, add root keys to generated YAML, use vault references instead of passwords

swagner-de:


networking_ccloud/tools/netbox_config_gen.py
```diff
@@ -400,12 +407,35 @@ def main():
...
+    if not (args.switch_password or args.vault_ref):
+        parser.error("Either'--vault-ref' or '--switch-passsword' must be set")
+
+    if args.vault_ref:
+        args.switch_password = VAULT_REF_REPLACEMENT
+        m = re.match(r'^(\S+)\W*:\W*(\S+)', args.vault_ref)
+        if m:
+            vault_path, vault_field = m.group(1), m.group(2)
+        else:
+            parser.error(f'Invalid vault reference should be <path>:<field>, got {args.vault_ref}')
```
sebageek, line 432: You were discussing with @joker-at-work yesterday if you'd use this one or
```
if not m:
    errro()
do_something()
```
Did you want to change this or did we want to keep it this way? (both fine)

networking_ccloud/tools/netbox_config_gen.py
```diff
@@ -414,7 +444,15 @@ def main():
 
     if args.output:
         conf_data = cfg.dict(exclude_unset=True, exclude_defaults=True, exclude_none=True)
+
+        if args.wrap_in:
+            for k in reversed(args.wrap_in.split('/')):
+                conf_data = {k: conf_data}
```
sebageek, line 450: Nice

networking_ccloud/tools/netbox_config_gen.py
```diff
@@ -400,12 +407,35 @@ def main():
...
     args = parser.parse_args()
 
+    if args.switch_password and args.vault_ref:
+        parser.error("You may only use one of '--vault-ref' or '--switch-passsword'")
+    if not (args.switch_password or args.vault_ref):
+        parser.error("Either'--vault-ref' or '--switch-passsword' must be set")
+
+    if args.vault_ref:
+        args.switch_password = VAULT_REF_REPLACEMENT
+        m = re.match(r'^(\S+)\W*:\W*(\S+)', args.vault_ref)
```
sebageek, line 428: Could add names to the catchgroups (`?P<path>`, `?P<field>`)

networking_ccloud/tools/netbox_config_gen.py
```diff
@@ -400,12 +407,35 @@ def main():
...
+        if m:
+            vault_path, vault_field = m.group(1), m.group(2)
+        else:
+            parser.error(f'Invalid vault reference should be <path>:<field>, got {args.vault_ref}')
+
+    if args.wrap_in:
+        m = re.match(r'^(\w+\/)*(\w+)$', args.wrap_in)
```
sebageek, line 435: I was a bit confused by the two (), as I thought you wanted to grab something out of the string. How abut adding `?:` if you just want to use them only for grouping? Also, why does the `/` need escaping?

How about `r'^(?:\w+/)*\w+$'`?

sebageek (commented):
Somewhat pedantic review, all comments up for discussion. ;)

networking_ccloud/tools/netbox_config_gen.py
swagner-de: No, JOker was suggesting to change it, but i like the `parser.error`-way.

networking_ccloud/tools/netbox_config_gen.py
```diff
@@ -31,11 +31,13 @@
 
 LOG = logging.getLogger(__name__)
 
-SWITCHGROUP_ROLE_VPOD = 'vpod'
-SWITCHGROUP_ROLE_STPOD = 'stpod'
-SWITCHGROUP_ROLE_APOD = 'apod'
-SWITCHGROUP_ROLE_NETPOD = 'netpod'
-SWITCHGROUP_ROLE_BPOD = 'bpod'
+SWITCHGROUP_ROLE_VPOD = 'bb'
+SWITCHGROUP_ROLE_BPOD = 'bb'
```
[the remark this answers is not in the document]
swagner-de: yes, but we are treating and prefixing them the same, so I saw no reason to not assign it the same const value.

networking_ccloud/tools/netbox_config_gen.py
swagner-de: but its just to and I did not see a large benefit here, except the regex getting longer.

networking_ccloud/tools/netbox_config_gen.py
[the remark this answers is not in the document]
sebageek: With the legacy infrastructure we have bpods that have overlapping numbers with bpods (at least I think we do). Won't we need handle them as well?

networking_ccloud/tools/netbox_config_gen.py
sebageek: It's documenting the regex. With one view of the regex you know what these () are meant for.

networking_ccloud/tools/netbox_config_gen.py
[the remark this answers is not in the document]
swagner-de: But without any idea how they will be connected in the future? At the moment, we cannot grab them as long as they are not directly connected to a EVPN-Leaf. So I am reluctant to think about that now.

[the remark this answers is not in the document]
swagner-de: I have thus removed bpod handling alltogether.

networking_ccloud/tools/netbox_config_gen.py
```diff
@@ -400,12 +405,35 @@ def main():
...
+    if args.vault_ref:
+        args.switch_password = VAULT_REF_REPLACEMENT
+        m = re.match(r'^(?P<path>\S+)\W*:\W*(?P<field>\S+)', args.vault_ref)
+        if m:
+            vault_path, vault_field = m.group('path'), m.group('field')
+        else:
+            parser.error(f'Invalid vault reference should be <path>:<field>, got {args.vault_ref}')
+
+    if args.wrap_in:
+        m = re.match(r'^(?:\w+/)*(?:\w+)$', args.wrap_in)
```
sebageek, line 433: Just FYI you don't need that second pair of `()`.

sebageek (approved):

PR #38, opened 2022-05-18
Add vlan ranges to driver config

sebageek:
GlobalConfig now has a setting to define the default vlan ranges. These
can be overridden on a per-switchgroup basis.

This setting is not used inside the driver code itself yet, but can be
used by external tools (e.g. helm-charts) to generate the
network_vlan_ranges setting for OpenStack's VlanTypeDriver.


networking_ccloud/common/config/config_driver.py
```diff
@@ -43,6 +43,17 @@ def validate_ip_address(addr: str) -> str:
...
 
 
+def validate_vlan_ranges(vlan_range: str) -> str:
+    m = re.match(r"^(?P<start>\d+):(?P<end>\d+)$", vlan_range)
+    if not m:
+        raise ValueError(f"Vlan range '{vlan_range}' is not in format $start_num:$end_num")
+    if not (2 <= int(m.group('start')) <= 4095 and 2 <= int(m.group('end')) <= 4095):
+        raise ValueError(f"Both parts of '{vlan_range}' need to be in range of [2, 4095]")
+    if int(m.group('start')) >= int(m.group('end')):
+        raise ValueError(f"Vlan range '{vlan_range}' needs to have a start that is lower than its end")
```
joker-at-work, line 53: Why can't it be the same if we just need that one VLAN? Is it because the `end` is excluded? i.e. `2:3` => `2` and not `2, 3`?

sebageek: You are right, the same VLAN should be possible.

swagner-de:
Shouldn't we also add the override possibility on a per switchgroup basis to this PR?

PR #39, opened 2022-05-19
More filtering, sorting, interconnect port-channels for config generator

swagner-de:


networking_ccloud/tools/netbox_config_gen.py
```diff
@@ -36,6 +36,8 @@
 SWITCHGROUP_ROLE_APOD = 'ap'
 SWITCHGROUP_ROLE_NETPOD = 'np'
 
+DEFAULT_VLAN_RANGES = ["2000:3750"]
```
sebageek, line 39: Sounds a bit brutal to completely hardcode this. Should we maybe make this configurable and set this as default?
I acknowledge that all of this is pretty much tailored to our needs, but until now we tried to keep things in a way that somebody else could take this as a base, throw in their own values and run it. Maybe it's time for me to part ways with this idea - at least with the config generator.

networking_ccloud/tools/netbox_config_gen.py
```diff
@@ -80,21 +83,45 @@ def __init__(self, region, args, verbose=False, verify_ssl=False):
             self.netbox.http_session = requests.Session()
             self.netbox.http_session.verify = False
 
-    def _ensure_single_item(self, item_set: Iterable[Any], entity: str, identifier: str, attribute: str) -> Any:
+    @classmethod
+    def _ignore_filter(cls, items: Iterable[NbRecord]) -> Generator[NbRecord, None, None]:
+        for item in items:
+            if any((x.slug in cls.ignore_tags for x in getattr(item, 'tags', list()))):
```
sebageek: You can remove the `()` inside the `any()`, the generator expression works perfectly fine inside it like this: `any(x for x in y)`

networking_ccloud/tools/netbox_config_gen.py
```diff
@@ -19,7 +19,7 @@
 from operator import itemgetter
 import re
 import urllib3
-from typing import Any, Dict, Optional, List, Tuple, Set, Iterable
+from typing import Any, Dict, Optional, List, Tuple, Set, Iterable, Generator, Union
```
sebageek: sort alphabetically

networking_ccloud/tools/netbox_config_gen.py
```diff
@@ -253,20 +266,7 @@ def get_connected_devices(self, switches: List[NbRecord]) -> Tuple[Set[NbRecord]
...
-                    if lag:
-                        lag.members.append(iface.name)
-                    else:
-                        device_ports_map[far_device].append(conf.SwitchPort(switch=switch.name, lacp=True,
-                                                                            name=iface.lag.name, members=[iface.name]))
-                else:
-                    # not a lag
-                    device_ports_map[far_device].append(conf.SwitchPort(switch=switch.name, name=iface.name))
-        hgs = [conf.Hostgroup(binding_hosts=[h.name], direct_binding=True, members=m)
+        hgs = [conf.Hostgroup(binding_hosts=[h.name], direct_binding=True, members=self.sort_switchports(m))
```
sebageek, line 296: As you sort the switchports, should we sort Hostgroups as well? Sorting criteria would be a bit harded, but I'd go either with physnet name (this was what we went with on ACI, but we would need to find out the name by crossreferencing it; could be done once the whole config is assembled) or we sort it by the alphabetically first binding host of each group.

networking_ccloud/tools/netbox_config_gen.py
```diff
@@ -381,7 +383,8 @@ def generate_config(self):
                                                                               status='active')))
         interconnect_hostgroups = self.get_interconnect_hostgroups(nb_switches)
         asn_region = self.get_asn_region(self.region)
-        switchgroups = self.get_switchgroups(asn_region, nb_switches, switch_user, switch_password)
+        switchgroups = sorted(self.get_switchgroups(asn_region, nb_switches, switch_user, switch_password),
+                              key=lambda x: x.name)
```
sebageek, line 429: Are the switches inside a switchgroup already sorted?

networking_ccloud/tools/netbox_config_gen.py
```diff
@@ -224,6 +236,7 @@ def get_switchgroups(self, asn_region: int, switches: List[NbRecord],
             # FIXME: Warn when a switchgroup is larger 2
             # we derive loopback1/asn from switch_number_resources which we already grouped on
             # so a switchgroup is guranteed to have the same values
+            members = sorted(members, key=lambda x: x.name)
```
sebageek: you can also use inplace sort here, i.e. `members.sort(key=...)`

networking_ccloud/tools/netbox_config_gen.py
```diff
@@ -370,6 +370,8 @@ def get_metagroups(self, connected_devices: Set[NbRecord]) -> List[conf.Hostgrou
         for name, count in Counter(cluster_names).items():
             if count > 1:
                 raise ConfigException(f'Cluster {name} appeared {count} times')
+        for name, metagroup in metagroups.items():
+            metagroup.members = list(sorted(metagroup.members))  # type: ignore
```
sebageek: You can use inplace sorting here, `metagroup.members.sort()`. Btw, sorted() should return a list already, not a generator or something else.

networking_ccloud/tools/netbox_config_gen.py
```diff
@@ -134,6 +134,18 @@ def parse_ccloud_switch_number_resources(cls, device_name: str) -> Dict[str, int
...
+    def sort_switchports(cls, switchports: List[conf.SwitchPort]) -> List[conf.SwitchPort]:
+
+        def keyfunc(sp: conf.SwitchPort) -> Tuple[Union[int, str]]:
+            if sp.name:
+                tokens = re.split(r"(\d+)", sp.name)
+                return tuple(int(x) if x.isdecimal() else x for x in tokens)
+            else:
+                return tuple(sp.switch)
+
+        return list(sorted(switchports, key=keyfunc))
```
sebageek: Already mentioned below, `sorted()` returns a list already.

networking_ccloud/tools/netbox_config_gen.py
```diff
@@ -80,21 +83,45 @@ def __init__(self, region, args, verbose=False, verify_ssl=False):
             self.netbox.http_session = requests.Session()
             self.netbox.http_session.verify = False
 
-    def _ensure_single_item(self, item_set: Iterable[Any], entity: str, identifier: str, attribute: str) -> Any:
+    @classmethod
+    def _ignore_filter(cls, items: Iterable[NbRecord]) -> Generator[NbRecord, None, None]:
```
sebageek, line 87: Why does this need to be a generator? Is that only so you don't need to fetch certain attributes twice and you can't make a list out of them beforehand? It makes the code flow a bit unintuitive to me in some places.

networking_ccloud/tools/netbox_config_gen.py
swagner-de: Where would you suggest to feed this in? An argument is a  bit to unstatic to me. The config generator is anyhow tightly tailored to our netbox, I cannot think that anybody else can use it without actually seeing how netbox looks for us.

networking_ccloud/tools/netbox_config_gen.py
swagner-de: I made a fixup commit for this. I create the HGs individually. Metagroup-HGs have anyhow just a single binding host. They will be ordered by `binding_hosts[0]` and all their direct binding HGs are before them in alphabetical order. We also have interconnect HGs which also only have exactly one binding host. Then, and that is just temporary (I assume), we have some leftover direct binding hosts (asrs, apods) that are direct binding and thus also only have one binding host. So, yes, I'd like to use the binding host option

networking_ccloud/tools/netbox_config_gen.py
swagner-de: https://github.com/sapcc/networking-ccloud/pull/39/files#diff-b211a5bc1049300b3d41501d74797736c0411808a9fa4829d026ffba41a6e515R263
Yes, they are.

networking_ccloud/tools/netbox_config_gen.py
swagner-de: This is likely going to take a `RecordSet` from `pynetbox` which is a generator already. I did not want to break with that concept as in some places, the netbox responses generator is taken and iterated over straight away. If I make this return a list, I break that generator concept. I know it's just minor memory allocation here and there, but still I would like to keep it.

networking_ccloud/tools/netbox_config_gen.py
sebageek: Are we yet at the point where we need a config for the config generator? ;) Maybe we are. Initially I thought about having this as a argument in `argparse`. But you might be right with that if somebody else would want to use it they'd have to write their own config generator (good thing we have most of the sanity checks inside the DriverConfig as validators). Let's keep it this way then.

swagner-de: On that note, the netbox url is also hardcoded already :|

sebageek: That settles that :D

networking_ccloud/tools/netbox_config_gen.py
sebageek: Okay, then let's keep it for now. We can still revisit this if we see the need for it.

sebageek (approved):

PR #42, opened 2022-05-24
Explicitly list AZs in driver config

sebageek:
In the future each AZ will need a number and a suffix. As we don't want
to require a specific AZ format, we allow this to be configured. This
also allows us a somewhat easier AZ handling, as we don't need to
iterate over all switchgroups and could also define AZs for which we
don't have a switchgroup.

Adaption of the config generator for this is not part of this commit.

### DIFF
--- a/networking_ccloud/tools/netbox_config_gen.py
+++ b/networking_ccloud/tools/netbox_config_gen.py
@@ -14,9 +14,9 @@
 
 from collections import defaultdict, Counter
 import ipaddress
-from itertools import groupby
+from itertools import chain, groupby
 import logging
-from operator import itemgetter
+from operator import attrgetter, itemgetter
 import re
 import urllib3
 from typing import Any, Dict, Generator, Iterable, List, Optional, Set, Tuple, Union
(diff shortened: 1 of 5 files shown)


networking_ccloud/tools/netbox_config_gen.py
```diff
@@ -14,7 +14,7 @@
 
 from collections import defaultdict, Counter
 import ipaddress
-from itertools import groupby
+from itertools import groupby, chain
```
sebageek: could™ be sorted alphabetically

networking_ccloud/tools/netbox_config_gen.py
```diff
@@ -114,6 +114,23 @@ def _ensure_single_item(cls, item_set: Iterable[Any], entity: str, identifier: s
                                         f'Unexpected values for {attribute}, expected 1 but got {len(item_set)}')
         return item_set.pop()
 
+    def get_azs(self) -> List[conf.AvailabilityZone]:
+        azs = list()
+        for site in self.netbox.dcim.sites.filter(region=self.region):
+            suffix = site.slug[len(self.region):].lower()
+            number = ord(suffix) - ord('a') + 1
+            azs.append(conf.AvailabilityZone(name=site.slug, suffix=suffix, number=number))
+        return sorted(azs, key=lambda x: x.number)
```
sebageek: `key=operator.attrgetter('number')` if you want

networking_ccloud/tools/netbox_config_gen.py
```diff
@@ -114,6 +114,23 @@ def _ensure_single_item(cls, item_set: Iterable[Any], entity: str, identifier: s
...
+
+    def get_cloud_vrfs(self) -> List[conf.VRF]:
+        vrfs = list()
+        search_strings = ['CC-CLOUD', 'CC-MGMT']
+        nb_return = chain(*(self.netbox.ipam.vrfs.filter(q=x, tenant=self.connection_tenants) for x in search_strings))
+        for vrf in nb_return:
+            rd_suffix = int(vrf.rd[vrf.rd.find(':') + 1:])
```
sebageek, line 130: `int(vrf.rd.split(':', 1)[1]))`

networking_ccloud/tools/netbox_config_gen.py
```diff
@@ -114,6 +114,23 @@ def _ensure_single_item(cls, item_set: Iterable[Any], entity: str, identifier: s
...
+            vrfs.append(conf.VRF(name=vrf.name, number=rd_suffix))  # type: ignore
+        return sorted(vrfs, key=lambda x: x.number)
```
sebageek: `key=operator.attrgetter('number')` if you want (maybe also `from operator import attrgetter`)

networking_ccloud/tools/netbox_config_gen.py
swagner-de: Does this make a difference?

networking_ccloud/tools/netbox_config_gen.py
swagner-de: okay, let's import `attrgetter` then.

networking_ccloud/tools/netbox_config_gen.py
sebageek: I think it's more readable. Splitting the thing at `":"` and then using the second half was more intuitive for me than doing the split manually with a find, slicing and "offset calculation".

PR #43, opened 2022-05-24
Explicitly add RTs to agent message format

sebageek:
We now put the Route Targets that should be exported/imported directly
into the agent message / SwitchConfig format. This means that we'll have
less business logic inside the agent and more in the neutron-server part
of the driver (which is a good thing, as we will have multiple agents
and therefore will not need to duplicate business logic). It will also
help when we implement the config diff functionality, as we can then
just fill the SwitchConfig object with what's on the switch and diff it
with the "should" version from OpenStack.

### DIFF
--- a/networking_ccloud/ml2/agent/common/messages.py
+++ b/networking_ccloud/ml2/agent/common/messages.py
@@ -211,13 +211,28 @@ class SwitchConfigUpdate(pydantic.BaseModel):
 
         return iface
 
-    def get_or_create_iface(self, switchport):
+    def get_iface(self, name):
         if self.ifaces is None:
             self.ifaces = []
 
         for iface in self.ifaces:
-            if iface.name == switchport.name:
+            if iface.name == name:
                 return iface
+        return None
+
+    def get_or_create_iface(self, name):
+        iface = self.get_iface(name)
+        if iface:
+            return iface
+
+        iface = IfaceConfig(name=name)
+        self.ifaces.append(iface)
+        return iface
+
+    def get_or_create_iface_from_switchport(self, switchport):
+        iface = self.get_iface(switchport.name)
+        if iface:
+            return iface
 
         iface = IfaceConfig.from_switchport(switchport)
         self.ifaces.append(iface)
@@ -278,7 +293,7 @@ class SwitchConfigUpdateList:
             # interface config
             if not is_bgw:
                 for sp in switchports:
-                    iface = scu.get_or_create_iface(sp)
+                    iface = scu.get_or_create_iface_from_switchport(sp)
                     iface.add_trunk_vlan(seg_vlan)
 
                     if hg_config.direct_binding and not hg_config.role:
--- a/networking_ccloud/ml2/agent/common/messages.py
+++ b/networking_ccloud/ml2/agent/common/messages.py
@@ -13,6 +13,8 @@
 #    under the License.
 
 from enum import Enum
+import ipaddress
+import re
 from typing import List
 
 from neutron_lib.plugins.ml2 import api as ml2_api
@@ -24,6 +26,52 @@ from networking_ccloud.common import constants as cc_const
 from networking_ccloud.ml2.agent.common.api import CCFabricSwitchAgentRPCClient
 
 LOG = logging.getLogger(__name__)
+
+
+def validate_route_target(rt) -> str:
+    # https://datatracker.ietf.org/doc/html/rfc4360#section-4
+    # https://datatracker.ietf.org/doc/html/rfc4364#section-4.2
+    if not isinstance(rt, str):
+        rt = str(rt)
+
+    if rt.isdecimal():
+        rt = int(rt)
+        if (rt >> 48) & 0xFF != 0x2:
+            raise ValueError("Route targets are extended attributes and need to have a 0x2 in the type field's "
+                             f"low order position - {rt} does not have this set (RFC 4360 sec 4)")
+        rt_type = rt >> 56
+        # 8byte rt
+        if rt_type == 0:
+            # 2 byte asn, 4byte admin field
+            a = (rt >> 32) & (2**16 - 1)
+            b = rt & (2**32 - 1)
+            rt = f"{a}:{b}"
+        elif rt_type == 1:
+            # 4 byte ip, 2byte admin field
+            a = (rt >> 16) & (2**32 - 1)
+            b = rt & (2**16 - 1)
+            rt = f"{ipaddress.ip_address(a)}:{b}"
+        elif rt_type == 2:
+            # 4 byte asn, 2byte admin field
+            a = (rt >> 16) & (2**32 - 1)
+            b = rt & (2**16 - 1)
+            if a > 2**16:
+                rt = f"{a >> 16}.{a & (2**16 - 1)}:{b}"
+            else:
+                rt = f"{a}:{b}"
+        else:
+            raise ValueError(f"Rt {rt} has invalid type {rt_type}, see RFC 4364 sec 4.2")
+    else:
+        # ip:num, asn:num, num:num
+        m = re.match(r"^(?P<first>(?:\d+\.\d+\.\d+\.\d+)|(?:\d+\.\d+)|(?:\d+)):(?P<second>\d+)$", rt)
+        if not m:
+            raise ValueError(f"RT '{rt}' is not in a recognizable format")
+        # automatically reformat cases where ASN is > 16bit to AS dot notation
+        first = m.group('first')
+        if first.isdecimal() and int(first) >= 2**16:
+            first = int(first)
+            rt = f"{first >> 16}.{first & (2**16 - 1)}:{m.group('second')}"
+    return rt
 
 
 class OperationEnum(str, Enum):
@@ -45,15 +93,24 @@ class VXLANMapping(pydantic.BaseModel):
 class BGPVlan(pydantic.BaseModel):
     # FIXME: validator
     rd: str
+    rd_evpn_domain_all: bool = False
     vlan: pydantic.conint(gt=0, lt=4094)
-    vni: pydantic.conint(gt=0, lt=2**24)
-    bgw_mode: bool = False
+
+    rt_imports: List[str] = []
+    rt_exports: List[str] = []
+    rt_imports_evpn: List[str] = []
+    rt_exports_evpn: List[str] = []
+
+    _norm_rt_imports = pydantic.validator('rt_imports', each_item=True, allow_reuse=True)(validate_route_target)
+    _norm_rt_exports = pydantic.validator('rt_exports', each_item=True, allow_reuse=True)(validate_route_target)
+    _norm_rt_imports_evpn = pydantic.validator('rt_imports_evpn',
+                                               each_item=True, allow_reuse=True)(validate_route_target)
+    _norm_rt_exports_evpn = pydantic.validator('rt_exports_evpn',
+                                               each_item=True, allow_reuse=True)(validate_route_target)
 
 
 class BGP(pydantic.BaseModel):
     asn: str
-
-    # regional asn (only used for bgws)
     asn_region: str
 
     vlans: List[BGPVlan] = None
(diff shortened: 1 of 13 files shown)


networking_ccloud/ml2/agent/common/messages.py
```diff
@@ -26,6 +28,52 @@
...
+def validate_route_target(rt) -> str:
+    # https://datatracker.ietf.org/doc/html/rfc4360#section-4
+    # https://datatracker.ietf.org/doc/html/rfc4364#section-4.2
+    if not isinstance(rt, str):
+        rt = str(rt)
+
+    if rt.isnumeric():
+        rt = int(rt)
+        if (rt >> 48) & 0xFF != 0x2:
+            raise ValueError("Route targets are extended attributes and need to have a 0x2 in their low order position "
```
swagner-de: `and need to have a 0x2 in the type field's low order position` maybe?

networking_ccloud/ml2/agent/common/messages.py
```diff
@@ -26,6 +28,52 @@
...
+                             f"- {rt} does not have this set (RFC 4360 sec 4)")
+        rt_type = rt >> 56
+        # 8byte rt
+        if rt_type == 0:
+            # 2 byte asn, 4byte number
+            a = (rt >> 32) & (2**16 - 1)
+            b = rt & (2**32 - 1)
```
swagner-de, line 47: Maybe call this `admin_field` in accordance with the RFC?

networking_ccloud/ml2/agent/common/messages.py
```diff
@@ -45,29 +93,49 @@ class VXLANMapping(pydantic.BaseModel):
...
+    def add_vlan(self, vlan, vni, bgw_mode=False):
+        # FIXME: raise if vni > 2byte (can't encode it in RT otherwise, write snarky commit message for that)
         if not self.vlans:
             self.vlans = []
+
+        rd = f"{self.asn}:{vni}"
         for bv in self.vlans:
-            if bv.rd == rd and bv.vlan == vlan and bv.vni == vni:
+            if bv.rd == rd and bv.vlan == vlan:
                 return
```
swagner-de, line 129: I am not sure if validators are run everytime an objects property is changed, but if so, couldn't that be a validator?
Also, is it a good idea to silently stop? Is it expected that we add VLANs to a BGP object that it is not responsible for?

networking_ccloud/ml2/agent/common/messages.py
```diff
@@ -26,6 +28,52 @@
...
+            b = rt & (2**16 - 1)
+            if a > 2**16:
+                rt = f"{a >> 16}.{a & (2**16 -1)}:{b}"
+            else:
+                rt = f"{a}:{b}"
+        else:
+            raise ValueError(f"Rt {rt} has invalid type {rt_type}, see RFC 4364 sec 4.2")
+    else:
+        # ip:num, asn:num, num:num
+        m = re.match(r"^(?P<first>(?:\d+\d+\.\d+\.\d+)|(?:\d+\.\d+)|(?:\d+)):(?P<second>\d+)$", rt)
```
joker-at-work: Why `\d+\d+`? Doesn't the first `\d+` already contain the second?

networking_ccloud/ml2/agent/common/messages.py
```diff
@@ -26,6 +28,52 @@
...
+        if not m:
+            raise ValueError("RT is not in a recognizable format")
```
joker-at-work: Do we maybe want to know which RT here?

networking_ccloud/ml2/agent/common/messages.py
```diff
@@ -26,6 +28,52 @@
...
+        # automatically reformat cases where asn is > 16bit to as dot notation
```
joker-at-work: `to as` -> `as` or `to` only?

networking_ccloud/ml2/agent/common/messages.py
```diff
@@ -26,6 +28,52 @@
...
+        first = m.group('first')
+        if first.isnumeric() and int(first) >= 2**16:
```
joker-at-work: You might want `isdecimal()` instead.

networking_ccloud/ml2/agent/common/messages.py
```diff
@@ -26,6 +28,52 @@
...
+            first = int(first)
+            rt = f"{first >> 16}.{first & (2**16 -1)}:{m.group('second')}"
```
joker-at-work: is there a space missing before `1` and after `-`?

networking_ccloud/tools/agent_rpc_caller.py
```diff
@@ -58,10 +60,16 @@ def main():
         topic = cc_const.SWITCH_AGENT_TOPIC_MAP[CONF.agent]
         client = CCFabricSwitchAgentRPCClient(topic)
 
+    try:
+        kwargs = json.loads(CONF.kwargs)
+    except json.decoder.JSONDecodeError as e:
+        print(f"Error: Could not decode json in kwargs: {e}")
+        sys.exit(1)
+
     print(f"Doing RPC call {CONF.method}(*{CONF.args}) with topic {topic}")
```
joker-at-work: Does this not need the kwargs? Imho, it now shows incomplete information about the call that was being made.

networking_ccloud/ml2/agent/common/messages.py
```diff
@@ -211,13 +211,28 @@ def add_iface(self, iface):
...
+        return None
+
+    def get_or_create_iface(self, name):
+        iface = self.get_iface(name)
+        if iface:
+            return iface
+
+        iface = IfaceConfig(name=name)
+        self.ifaces.append(iface)
+        return iface
```
joker-at-work, line 230: Could this be racy by being called from different threads/greenthreads?

networking_ccloud/ml2/agent/common/messages.py
```diff
@@ -211,13 +211,28 @@ def add_iface(self, iface):
 
         return iface
 
-    def get_or_create_iface(self, switchport):
+    def get_iface(self, name):
         if self.ifaces is None:
             self.ifaces = []
```
joker-at-work, line 216: Is order important or could this be a dict for faster access and for not having duplicates? Looks like `.name` is key anyways.

networking_ccloud/ml2/agent/eos/switch.py
```diff
@@ -236,6 +242,72 @@ def get_removable_translations_cmds(iface_name, wanted_translations):
...
+    def get_config(self):
+        # get infos from the device for everything that we have a model for
+        config = agent_msg.SwitchConfigUpdate(switch_name=self.name, operation=Op.add)
+
+        # vlans
+        for vlan, data in self.send_cmd("show vlan")['vlans'].items():
+            config.add_vlan(int(vlan), data['name'])
+
+        # vxlan_maps
+        vxlan_maps = self.send_cmd("show interfaces Vxlan1")['interfaces']['Vxlan1']['vlanToVniMap']
```
joker-at-work, line 268: `Vxlan1` is hardcoded? O.o

networking_ccloud/ml2/agent/eos/switch.py
```diff
@@ -236,6 +242,72 @@ def get_removable_translations_cmds(iface_name, wanted_translations):
...
+
+        # bgp
+        bgp_sum = self.send_cmd("show bgp summary")['vrfs']['default']
+        config.bgp = agent_msg.BGP(asn=bgp_sum['asn'], asn_region=self.drv_conf.global_config.asn_region, vlans=[])
+        curr_bgp_vlans = self.send_cmd("show bgp evpn instance")['bgpEvpnInstances']
+        vre = re.compile(r"VLAN (?P<vlan>\d+)")
+        for vlan_name, data in curr_bgp_vlans.items():
+            m = vre.match(vlan_name)
+            if not m:
+                LOG.warning("Could not match bgp vpn instanec name '%s'", vlan_name)
```
joker-at-work: `instanec` -> `instance`

networking_ccloud/ml2/agent/eos/switch.py
```diff
@@ -236,6 +242,72 @@ def get_removable_translations_cmds(iface_name, wanted_translations):
...
+            config.bgp.vlans.append(bv)
+            # FIXME: redistribute learned flag?
+
+        # ifaces - vlans
+        ifaces_vlans = self.send_cmd("show interfaces vlans")['interfaces']
+        for name, data in ifaces_vlans.items():
+            if name == "Vxlan1":
+                continue
+            iface = config.get_or_create_iface(name)
+            if data.get("untaggedVlan", 1) != 1:
```
joker-at-work, line 299: Can this be `'1'`? We converted a lot to `int()` above, so I'm curious why this comes back as `int` already.

networking_ccloud/ml2/agent/eos/switch.py
```diff
@@ -113,13 +113,27 @@ def _make_config_from_update(self, config):
...
-                wanted_maps = [(v.vlan, v.vni) for v in config.vxlan_maps]
                 curr_maps = self.send_cmd("show interfaces Vxlan1")['interfaces']['Vxlan1']['vlanToVniMap']
-                for vlan, data in curr_maps.items():
-                    if data.get('vni') and (int(vlan), data['vni']) not in wanted_maps:
-                        commands.append(f"no vxlan vlan {vlan} vni {data['vni']}")
+                if config.operation == Op.add:
+                    # purge mappings referenced by config update, but point to something else
+                    for vlan, data in curr_maps.items():
+                        switch_vlan = int(vlan)
+                        switch_vni = data['vni']
```
joker-at-work, line 124: We use `data.get('vni')` below, why are we sure it exists here?

networking_ccloud/ml2/agent/common/messages.py
sebageek: The idea here is that if we construct the config update and due "some sort of codeflow" it might happen that we add the same vlan mapping twice. I want to make sure we don't end up in config generation with something that tries to add the same thing twice and then fails.

Validators are not run on object modification:
```
In [37]: agent_msg.BGPVlan(rd="123", vlan=1234, rt_imports=["123"])
---------------------------------------------------------------------------
ValidationError                           Traceback (most recent call last)
Input In [37], in <module>
----> 1 agent_msg.BGPVlan(rd="123", vlan=1234, rt_imports=["123"])

File ~/venvs/fabric/lib/python3.8/site-packages/pydantic/main.py:331, in pydantic.main.BaseModel.__init__()

ValidationError: 1 validation error for BGPVlan
rt_imports -> 0
  Route targets are extended attributes and need to have a 0x2 in the type field's low order position - 123 does not have this set (RFC 4360 sec 4) (type=value_error)

In [38]: bv = agent_msg.BGPVlan(rd="123", vlan=1234)

In [39]: bv.rt_imports.append("123")

In [40]: 
```
Though calling validation is possible (at least it reads like it), see here: https://github.com/samuelcolvin/pydantic/issues/1864

networking_ccloud/ml2/agent/common/messages.py
sebageek: `as`  to `AS`

networking_ccloud/ml2/agent/common/messages.py
sebageek: These objects are generally created by a single thread and are not designed for two threads creating it at the same time. The thing would need a whole redesign for that.

networking_ccloud/ml2/agent/eos/switch.py
sebageek: `Vxlan1` is a special interface on the `EOS` platform which contains all the vxlan mappings and some other magical stuff. In theory™ there could be multiple Vxlan interfaces, but no EOS platform supports that yet to my knowedge (and our usecase wouldn't need it).

networking_ccloud/ml2/agent/eos/switch.py
sebageek: Values in the `EOS` json response are mostly an int where they should be an int. Dict keys on the other hand are strings, so they need int converting.

```
qa-de-1-sw4113a-bb271#show interfaces vlans | json | inc untagged
            "untaggedVlan": 2121
            "untaggedVlan": 2121
            "untaggedVlan": 2121
```

PR #44, opened 2022-05-25
Pull InfraNetworks from netbox and change driver config

swagner-de:
Pulling InfraNetworks from netbox using vlans assigned to the server
facing ports, as well as the SVIs modelled on the TORs for the anycast
gateteway.
Also made VNI an mandatory field, as we potentially want to stretch
every VLAN over TOR borders.
VRF is also mandtory if the network is L3 enabled.

### DIFF
--- a/networking_ccloud/common/config/config_driver.py
+++ b/networking_ccloud/common/config/config_driver.py
@@ -196,20 +196,42 @@ class SwitchPort(pydantic.BaseModel):
 
 class InfraNetwork(pydantic.BaseModel):
     name: str
-    native_vlan_pool: str = None
-    vlan: pydantic.conint(gt=0, lt=4096)
-    networks: List[str] = None
-    vni: pydantic.conint(gt=0, lt=2**24) = None
+    vlan: pydantic.conint(gt=1, lt=4095)
+    vrf: str = None
+    networks: List[str] = []
+    vni: pydantic.conint(gt=0, lt=2**24)
     untagged: bool = False
-    dhcp_relays: List[str] = None
+    dhcp_relays: List[str] = []
 
-    _normalize_networks = pydantic.validator('networks', each_item=True, allow_reuse=True)(validate_ip_address)
+    _normalize_relays = pydantic.validator('dhcp_relays', each_item=True, allow_reuse=True)(validate_ip_address)
+
+    def __hash__(self) -> int:
+        return hash((self.name, self.vlan, self.vrf, tuple(self.networks),
+                    self.vni, self.untagged, tuple(self.dhcp_relays)))
+
+    @pydantic.validator('networks', each_item=True)
+    def ensure_host_bit_set(cls, net):
+        net = ipaddress.ip_interface(net)
+        if str(net) == str(net.network):
+            raise ValueError(f'Network {net} is supposed to be used as gateway and hence needs hosts bits set')
+        return str(net)
 
     @pydantic.root_validator
     def ensure_correct_value_combination(cls, values):
-        # FIXME: we probably need a different logic here
-        if bool(values.get('vni')) ^ bool(values.get('networks')):
-            raise ValueError("If network is set vni needs to be set and vice versa")
+        if len(values.get('networks')) > 0 and not bool(values.get('vrf')):
+            raise ValueError("If network is given a VRF must be set too")
+        if len(values.get('dhcp_relays')) > 0 and not len(values.get('networks')) > 0:
+            raise ValueError("If dhcp_relays is given a network must be present too")
+        return values
+
+    @pydantic.root_validator
+    def ensure_dhcp_relay_not_in_networks(cls, values):
+        for network in values.get('networks', []):
+            network = ipaddress.ip_interface(network)
+            for relay in values.get('dhcp_relays', []):
+                relay = ipaddress.ip_address(relay)
+                if relay in network.network:
+                    raise ValueError(f'dhcp_relay {relay} is contained in network {network}')
         return values
 
 
--- a/networking_ccloud/common/config/config_driver.py
+++ b/networking_ccloud/common/config/config_driver.py
@@ -553,6 +553,22 @@ class DriverConfig(pydantic.BaseModel):
 
         return values
 
+    @pydantic.root_validator
+    def ensure_all_infra_network_vrf_exist(cls, values):
+        if 'global_config' not in values:
+            return values
+        if 'hostgroups' not in values:
+            return values
+        global_config:  GlobalConfig = values['global_config']
+        hgs: List[Hostgroup] = values['hostgroups']
+        vrf_names = set(x.name for x in global_config.vrfs)
+        for hg in hgs:
+            if hg.infra_networks:
+                for net in hg.infra_networks:
+                    if net.vrf and net.vrf not in vrf_names:
+                        raise ValueError(f'Associated VRF {net.vrf} of infra network {net.name} is not existing')
+        return values
+
     def get_platforms(self):
         """Get all platforms as a set used in the given config"""
         v = set()
--- a/networking_ccloud/common/config/config_driver.py
+++ b/networking_ccloud/common/config/config_driver.py
@@ -431,6 +431,24 @@ class GlobalConfig(pydantic.BaseModel):
     _normalize_vlan_ranges = pydantic.validator('default_vlan_ranges',
                                                 each_item=True, allow_reuse=True)(validate_vlan_ranges)
 
+    @pydantic.validator('vrfs')
+    def check_vrf_name_unique(cls, values):
+        names = set()
+        for vrf in values:
+            if vrf.name in names:
+                raise ValueError(f'VRF {vrf.name} is duplicated')
+            names.add(vrf.name)
+        return values
+
+    @pydantic.validator('vrfs')
+    def check_vrf_number_unique(cls, values):
+        nums = set()
+        for vrf in values:
+            if vrf.number in nums:
+                raise ValueError(f'VRF id {vrf.number} is duplicated on VRF {vrf.name}')
+            nums.add(vrf.number)
+        return values
+
 
 class DriverConfig(pydantic.BaseModel):
     global_config: GlobalConfig
--- a/networking_ccloud/tests/common/config_fixtures.py
+++ b/networking_ccloud/tests/common/config_fixtures.py
@@ -84,7 +84,7 @@ def make_hostgroups(switchgroup, nodes=10, ports_per_switch=2, offset=0, **kwarg
     for n in range(1, nodes + 1):
         ports = make_lacp(100 + n, switchgroup, ports_per_switch=ports_per_switch, offset=(n - 1) * ports_per_switch)
         binding_host = f"node{n:03d}-{switchgroup}"
-        hg = config_driver.Hostgroup(binding_hosts=[binding_host], members=ports)
+        hg = config_driver.Hostgroup(binding_hosts=[binding_host], members=ports, **kwargs)
         groups.append(hg)
     return groups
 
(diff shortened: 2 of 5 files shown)


networking_ccloud/common/config/config_driver.py
```diff
@@ -196,22 +196,43 @@ def check_port_name_in_lacp_mode(cls, values):
...
+    _normalize_relays = pydantic.validator('dhcp_relays', each_item=True, allow_reuse=True)(validate_ip_address)
+
+    def __hash__(self) -> int:
+        return hash((self.name, self.vlan, self.vrf, tuple(self.networks),
+                    self.vni, self.untagged, tuple(self.dhcp_relays)))
+
+    @pydantic.validator('networks', each_item=True)
+    def ensure_host_bit_set(cls, net):
+        net = ipaddress.IPv4Interface(net)
+        if int(net) & int(net.netmask) == int(net):
```
sebageek: I'm a big fan of bitwise ops, but here is a variant without it:
```python
if str(net) == str(net.network):
    ...
```

networking_ccloud/common/config/config_driver.py
```diff
@@ -196,22 +196,43 @@ def check_port_name_in_lacp_mode(cls, values):
...
-    _normalize_networks = pydantic.validator('networks', each_item=True, allow_reuse=True)(validate_ip_address)
+    _normalize_relays = pydantic.validator('dhcp_relays', each_item=True, allow_reuse=True)(validate_ip_address)
+
+    def __hash__(self) -> int:
+        return hash((self.name, self.vlan, self.vrf, tuple(self.networks),
+                    self.vni, self.untagged, tuple(self.dhcp_relays)))
+
+    @pydantic.validator('networks', each_item=True)
+    def ensure_host_bit_set(cls, net):
+        net = ipaddress.IPv4Interface(net)
```
sebageek: Why are we here explicitly using `IPv4Interface`, but later on (L230) using `ip_interface` in the validator?

networking_ccloud/common/config/config_driver.py
```diff
@@ -196,22 +196,43 @@ def check_port_name_in_lacp_mode(cls, values):
...
         return values
 
+    @pydantic.root_validator
+    def ensure_dhcp_relay_not_in_networks(cls, values):
+        for network in values.get('networks', []):
+            network = ipaddress.ip_interface(network)
+            for relay in values.get('dhcp_relays', []):
+                relay = ipaddress.ip_address(relay)
+                if relay in network.network:
+                    raise ValueError(f'dhcp_relay {str(relay)} is contained in network {network.with_prefixlen}')
```
sebageek: I think f-strings already call `str()` on objects and you should be fine not adding that as an extra here.

Also, `{network}` seems to be roughly the same as `{network.with_prefixlen}`, so we could use the first one for shorter code (but don't have to). See https://github.com/python/cpython/blob/main/Lib/ipaddress.py#L1397-L1399=

networking_ccloud/tools/netbox_config_gen.py
```diff
@@ -513,6 +596,12 @@ def main():
             for k in reversed(args.wrap_in.split('/')):
                 conf_data = {k: conf_data}
 
+        def ip_net_representer(dumper: yaml.SafeDumper, x):
+            return dumper.represent_scalar('tag:yaml.org,2002:str', x.with_prefixlen)
+
+        yaml.SafeDumper.add_representer(ipaddress.IPv4Interface, ip_net_representer)
+        yaml.SafeDumper.add_representer(ipaddress.IPv6Interface, ip_net_representer)
```
sebageek: Could you explain to me how this part works again? The config itself doesn't seem to have IPv?Interface as type somewhere. Are we explicitly using that as a type or not? Also, what about other parts where we might™ need to dump the config object? If we use this we maybe should write our own dumper or something - or add these two to some sort of helper method.

networking_ccloud/tools/netbox_config_gen.py
```diff
@@ -170,6 +171,56 @@ def handle_interface_or_portchannel(cls, iface: NbRecord, candidate_interfaces:
...
+
+        # FIXME: support untagged VLANs
+        # FIXME: support DHCP relay
+        for vlan in getattr(iface, 'tagged_vlans', list()):
+            mandatory_attrs = ('vid', 'tenant')
+            for attr in mandatory_attrs:
+                if not getattr(vlan, attr, None):
+                    raise ConfigException(f'VLAN {vlan.id} has no attribute {attr}')
+                if vlan.tenant.slug not in self.tenants:
+                    continue
```
sebageek: What does this `continue` do? Is it meant to continue the for loop in L191? I think you could replace the `continue` with a `pass` here and it would be the same codeflow (you could even remove the whole `if` beforehand).

networking_ccloud/tools/netbox_config_gen.py
```diff
@@ -170,6 +171,56 @@ def handle_interface_or_portchannel(cls, iface: NbRecord, candidate_interfaces:
...
+            if not getattr(vlan, 'group', None):
```
sebageek, line 215: `if not hasattr(vlan, 'group'):`

networking_ccloud/tools/netbox_config_gen.py
```diff
@@ -415,6 +476,30 @@ def get_metagroups(self, connected_devices: Set[NbRecord]) -> List[conf.Hostgrou
...
+            minimum_common_infra_nets.intersection_update(member.infra_networks)
+
+        # purge them from the member_hgs
+        for member in member_hgs:
+            remaining_infra_nets = list()
+            for infra_net in member.infra_networks:
+                if infra_net not in minimum_common_infra_nets:
+                    remaining_infra_nets.append(infra_net)
+            member.infra_networks = remaining_infra_nets
+        metagroup.infra_networks = sorted(minimum_common_infra_nets, key=lambda x: x.vlan)
```
sebageek: You could also use `key=operator.attrgetter('vlan')` here if you like.

networking_ccloud/tools/netbox_config_gen.py
```diff
@@ -415,6 +476,30 @@ def get_metagroups(self, connected_devices: Set[NbRecord]) -> List[conf.Hostgrou
...
+
+        return metagroup, member_hgs
```
sebageek: You use return here,but don't use the returned values later on. Any reason to return them if it's an in-place operation anyways?

networking_ccloud/common/config/config_driver.py
swagner-de: No reason. It's fixed

networking_ccloud/common/config/config_driver.py
swagner-de: I am a bit upset, cause I was actively looking for a less bit-ty way to do this and now it turns out it is this easy.

networking_ccloud/tools/netbox_config_gen.py
swagner-de: What if it has the attribute but `vlan.group == None`?

networking_ccloud/tools/netbox_config_gen.py
swagner-de: That block was indented too much.

networking_ccloud/tools/netbox_config_gen.py
swagner-de: I usually try to explicitly return even if the ops was in-place. I will consume the return values from the function, deal?

networking_ccloud/tools/netbox_config_gen.py
swagner-de: In an earlier draft I was using `ipaddress` objects, but removed them now and forgot removing this.

networking_ccloud/tools/netbox_config_gen.py
sebageek: Then mine would continue into the if and yours would not. Got it, let's keep yours.

networking_ccloud/tools/netbox_config_gen.py
sebageek: Hmm to be honest I think I like the way you have it, no need to consume the return values if it would be the same without them. I like references and in-place ops.

networking_ccloud/common/config/config_driver.py
```diff
@@ -196,20 +196,42 @@ def check_port_name_in_lacp_mode(cls, values):
 
 class InfraNetwork(pydantic.BaseModel):
     name: str
-    native_vlan_pool: str = None
-    vlan: pydantic.conint(gt=0, lt=4096)
-    networks: List[str] = None
-    vni: pydantic.conint(gt=0, lt=2**24) = None
+    vlan: pydantic.conint(gt=1, lt=4095)
+    vrf: str = None
```
sebageek, line 200: Now that you have rebased onto the VRF config branch - do we want to have a check in the driver config that all VRFs referenced by an InfraNetwork exists in our config? Would be an extra root validator on the DriverConfig class

networking_ccloud/tests/common/config_fixtures.py
```diff
@@ -128,10 +128,30 @@ def make_azs_from_switchgroups(switchgroups):
...
+    for i, name in enumerate(names):
+        vrfs.append(config_driver.VRF(name=name, number=i + 1))
+    return vrfs
+
+
+def make_vrfs_from_infra_nets(infra_nets):
+    if infra_nets is None:
+        return []
+
+    return make_vrfs(x.vrf for x in set(infra_nets) if x.vrf)
```
sebageek: shouldn't this be `make_vrfs(set(x.vrf for x in infra_nets if vrf))`?

networking_ccloud/tests/common/config_fixtures.py
```diff
@@ -128,10 +128,30 @@ def make_azs_from_switchgroups(switchgroups):
...
+
 # whole config
 def make_config(switchgroups=None, hostgroups=None, global_config=None):
     if not global_config:
-        global_config = make_global_config(availability_zones=make_azs_from_switchgroups(switchgroups))
+        infra_nets = list()
+        if hostgroups:
+            for hostgroup in hostgroups:
+                if hostgroup.infra_networks:
+                    infra_nets.extend(hostgroup.infra_networks)
```
sebageek: Could also make a `make_vrfs_from_hostgroups()` instead of `from_infra_nets`. Maybe that's more our use case?

networking_ccloud/common/config/config_driver.py
swagner-de: Yes, there is a commit doing exactly that.

networking_ccloud/tests/common/config_fixtures.py
swagner-de: Good idea.

networking_ccloud/common/config/config_driver.py
```diff
@@ -431,6 +431,29 @@ class GlobalConfig(pydantic.BaseModel):
...
+    def check_vrf_number_unique(cls, values):
+        if 'vrfs' not in values:
+            return values
+        nums = set()
+        for vrf in values['vrfs']:
+            if vrf.number in nums:
+                raise ValueError(f'VRF id {vrf.number} is duplicated on VRF {vrf.name}')
+            else:
+                nums.add(vrf.number)
+        return values
```
sebageek: If these two validators only do validation on the `vrf` attribute they don't need to be `root_validators`, do they? Could make them to be a validator only for `vrf`.

networking_ccloud/common/config/config_driver.py
```diff
@@ -431,6 +431,29 @@ class GlobalConfig(pydantic.BaseModel):
...
+    @pydantic.root_validator
+    def check_vrf_name_unique(cls, values):
+        if 'vrfs' not in values:
+            return values
+        names = set()
+        for vrf in values['vrfs']:
+            if vrf.name in names:
+                raise ValueError(f'VRF {vrf.name} is duplicated')
+            else:
+                names.add(vrf.name)
```
sebageek: You can remove the else and dedent `names.add()`

networking_ccloud/common/config/config_driver.py
```diff
@@ -431,6 +431,29 @@ class GlobalConfig(pydantic.BaseModel):
...
+    @pydantic.root_validator
+    def check_vrf_number_unique(cls, values):
+        if 'vrfs' not in values:
+            return values
+        nums = set()
+        for vrf in values['vrfs']:
+            if vrf.number in nums:
+                raise ValueError(f'VRF id {vrf.number} is duplicated on VRF {vrf.name}')
+            else:
+                nums.add(vrf.number)
```
sebageek: You can remove the else and dedent `names.add()`

networking_ccloud/common/config/config_driver.py
swagner-de: They do the validation on the list of VRFs, so they must be root validators, don't they?

networking_ccloud/tests/common/config_fixtures.py
```diff
@@ -128,10 +128,30 @@ def make_azs_from_switchgroups(switchgroups):
...
+def make_vrfs_from_hostgroups(hostgroups):
+    if hostgroups is None:
+        return []
+    infra_nets = list()
+    for hostgroup in hostgroups:
+        if hostgroup.infra_networks:
+            infra_nets.extend(hostgroup.infra_networks)
+    
+
+    return make_vrfs(set(x.vrf for x in set(infra_nets) if x.vrf))
```
sebageek: Do we still need `set(infra_nets)`?

networking_ccloud/common/config/config_driver.py
sebageek: The `vrfs` field is a single field / member of the class `GlobalConfig`. If you write a validator for it you can tell it to just validate `vrfs` and it will get the whole list (as long as you don't provide `each_item=True` to the validator).

PR #50, opened 2022-06-14
Reorganize read/write switchconfig methods into smaller methods

swagner-de:
Instead of reading and writing all configuration in on large method such
as `get_config` or `_make_config_from_update` we keep these methods but
move the generation of the config entities into smaller methods. Each
methods generates either configuration commands and reads config for a
singular top-level entity. This keeps the methods smaller and more
modular. We may also think about moving the top-level methods into the
base class later.

### DIFF
--- a/networking_ccloud/ml2/agent/common/messages.py
+++ b/networking_ccloud/ml2/agent/common/messages.py
@@ -84,10 +84,16 @@ class Vlan(pydantic.BaseModel):
     vlan: pydantic.conint(gt=0, lt=4094)
     name: str = None
 
+    def __lt__(self, other):
+        return self.vlan < other.vlan
+
 
 class VXLANMapping(pydantic.BaseModel):
     vni: pydantic.conint(gt=0, lt=2**24)
     vlan: pydantic.conint(gt=0, lt=4094)
+
+    def __lt__(self, other):
+        return self.vlan < other.vlan
 
 
 class BGPVlan(pydantic.BaseModel):
@@ -108,6 +114,9 @@ class BGPVlan(pydantic.BaseModel):
     _norm_rt_exports_evpn = pydantic.validator('rt_exports_evpn',
                                                each_item=True, allow_reuse=True)(validate_route_target)
 
+    def __lt__(self, other):
+        return self.vlan < other.vlan
+
 
 class BGP(pydantic.BaseModel):
     asn: str
@@ -117,6 +126,9 @@ class BGP(pydantic.BaseModel):
 
     _normalize_asn = pydantic.validator('asn', allow_reuse=True)(validate_asn)
     _normalize_asn_region = pydantic.validator('asn_region', allow_reuse=True)(validate_asn)
+
+    def sort(self):
+        self.vlans.sort()
 
     def add_vlan(self, vlan, vni, bgw_mode=False):
         # FIXME: raise if vni > 2byte (can't encode it in RT otherwise, write snarky commit message for that)
@@ -151,6 +163,9 @@ class IfaceConfig(pydantic.BaseModel):
     vlan_translations: List[VlanTranslation] = None
     portchannel_id: pydantic.conint(gt=0) = None
     members: List[str] = None
+
+    def __lt__(self, other):
+        return self.name < other.name
 
     @classmethod
     def from_switchport(cls, switchport):
@@ -187,6 +202,18 @@ class SwitchConfigUpdate(pydantic.BaseModel):
     @classmethod
     def make_object_from_net_data(self, vxlan_map, net_host_map):
         pass
+
+    def sort(self):
+        # in case we want to compare this or ship it to the user we may benefit from it being sorted
+        # also test cases would fail otherwise
+        if self.vlans:
+            self.vlans.sort()
+        if self.vxlan_maps:
+            self.vxlan_maps.sort()
+        if self.ifaces:
+            self.ifaces.sort()
+        if self.bgp:
+            self.bgp.sort()
 
     def add_vlan(self, vlan, name=None):
         if self.vlans is None:
--- a/networking_ccloud/ml2/agent/eos/switch.py
+++ b/networking_ccloud/ml2/agent/eos/switch.py
@@ -12,8 +12,11 @@
 #    License for the specific language governing permissions and limitations
 #    under the License.
 
+from collections import defaultdict
+from operator import attrgetter
 import re
 import time
+from typing import Dict, List, Optional
 
 from oslo_log import log as logging
 import pyeapi
@@ -25,16 +28,20 @@ from networking_ccloud.ml2.agent.common import messages as agent_msg
 from networking_ccloud.ml2.agent.common.messages import OperationEnum as Op
 from networking_ccloud.ml2.agent.common.switch import SwitchBase
 
+
 LOG = logging.getLogger(__name__)
 
 
 class EOSSwitch(SwitchBase):
+
+    CONFIGURE_ORDER = ['vlan', 'vxlan_mapping', 'bgp', 'ifaces']
+
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.drv_conf = get_driver_config()
 
     @classmethod
-    def get_platform(self):
+    def get_platform(cls):
         return cc_const.PLATFORM_EOS
 
     def login(self):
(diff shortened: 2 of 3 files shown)


networking_ccloud/ml2/agent/common/messages.py
```diff
@@ -188,6 +189,17 @@ class SwitchConfigUpdate(pydantic.BaseModel):
     def make_object_from_net_data(self, vxlan_map, net_host_map):
         pass
 
+    def dict(self, *args, **kwargs):
+        # in case we want to compare this or ship it to the user we may benefit from it being sorted
+        # also test cases would fail otherwise
```
sebageek: Sorting this is a very good idea and it is also something that we need for the diffing functionality.

Is it a good idea to sort this on `dict()`? This means we change the _model_, when we export it, where I'd normally would expect this to be a read-only operation. Also, we can't sort the thing without exporting it.

What do you think about putting this functionality into an extra method like `sort()`? We could then either let the user of this class decide when to sort or add a key to `dict()` with `sort_members=True|False` or `sort=True|False`.

networking_ccloud/ml2/agent/eos/switch.py
```diff
@@ -25,16 +27,21 @@
 from networking_ccloud.ml2.agent.common.messages import OperationEnum as Op
 from networking_ccloud.ml2.agent.common.switch import SwitchBase
 
+from typing import Dict, List, Optional
```
sebageek: I don't think pep8 has an extra block for typing in mind and these should go to the stdlib imports.

https://peps.python.org/pep-0008/#imports

networking_ccloud/ml2/agent/eos/switch.py
```diff
@@ -83,86 +90,119 @@ def get_switch_status(self):
...
-                wanted_vlans = [v.vlan for v in config.vlans]
-                device_vlans = self.send_cmd("show vlan")['vlans']
-                for device_vlan in map(int, device_vlans.keys()):
+    def get_vlan_config(self) -> List[agent_msg.Vlan]:
+        vlans = [agent_msg.Vlan(vlan=int(vlan), name=data['name'])
+                 for vlan, data in self.send_cmd("show vlan")['vlans'].items()]
+        return sorted(vlans, key=attrgetter('vlan'))
+
+    def _make_vlan_config(self, vlans: Optional[List[agent_msg.Vlan]], operation: Op) -> List[str]:
+        commands = list()
```
sebageek: While practically not a problem, I have to say I find these `foo = list()` calls somewhat aesthetically unpleasing in comparison to `foo = []`

networking_ccloud/ml2/agent/eos/switch.py
```diff
@@ -83,86 +90,119 @@ def get_switch_status(self):
...
+        if vlans:
+            if operation == Op.replace:
+                wanted_vlans = [v.vlan for v in vlans]
+                device_vlans = self.get_vlan_config()
+                for device_vlan in map(attrgetter('vlan'), device_vlans):
```
sebageek: Big fan of map, though the list is shorter. Your choice, though.

```python
map(attrgetter('vlan'), device_vlans)
[v.vlan for v in device_vlans]
```

networking_ccloud/ml2/agent/eos/switch.py
```diff
@@ -83,86 +90,119 @@ def get_switch_status(self):
...
-        if config.vxlan_maps is not None:
+    def get_vxlan_mapping(self) -> List[agent_msg.VXLANMapping]:
+        result = self.send_cmd("show interfaces Vxlan1")['interfaces']['Vxlan1']['vlanToVniMap']
+        vxlan_maps = [agent_msg.VXLANMapping(vni=int(data['vni']), vlan=int(vlan))
+                      for vlan, data in result.items() if 'vni' in data]
+        return sorted(vxlan_maps, key=attrgetter('vlan'))
+
+    def _make_vxlan_mapping_config(self, vxlan_maps: Optional[List[agent_msg.VXLANMapping]],
+                                   operation: Op) -> List[str]:
+        commands = list()
```
sebageek: `commands = []`

networking_ccloud/ml2/agent/eos/switch.py
```diff
@@ -83,86 +90,119 @@ def get_switch_status(self):
...
-                elif config.operation == Op.remove:
+                elif operation == Op.remove:
                     commands.append(f"no vxlan vlan {vmap.vlan} vni {vmap.vni}")
             commands.append("exit")
+        return commands
+
+    def get_bgp_vlan_config(self) -> List[agent_msg.BGPVlan]:
+        curr_bgp_vlans = self.send_cmd("show bgp evpn instance")['bgpEvpnInstances']
+        vre = re.compile(r"VLAN (?P<vlan>\d+)")
+        bgp_vlans = list()
```
sebageek: `list()`

networking_ccloud/ml2/agent/eos/switch.py
```diff
@@ -83,86 +90,119 @@ def get_switch_status(self):
...
+        bgp_sum = self.send_cmd("show bgp summary")['vrfs']['default']
+        bgp = agent_msg.BGP(asn=bgp_sum['asn'], asn_region=self.drv_conf.global_config.asn_region, vlans=[])
+        bgp.vlans = self.get_bgp_vlan_config()
+        return bgp
 
-        # bgp section
-        if config.bgp and config.bgp.vlans:
-            commands.append(f"router bgp {config.bgp.asn}")
+    def _make_bgp_config(self, bgp: Optional[agent_msg.BGP], operation: Op) -> List[str]:
+        commands = list()
```
sebageek: `list()`

networking_ccloud/ml2/agent/eos/switch.py
```diff
@@ -187,56 +227,107 @@ def _make_config_from_update(self, config):
...
+            iface = get_or_create_iface(name)
+            if data.get("untaggedVlan", 1) != 1:
+                iface.native_vlan = data['untaggedVlan']
+            iface.trunk_vlans = data.get('taggedVlans', [])
+
+        # vlan translations
+        vtrans = self.get_vlan_translations()
+        for name, translations in vtrans.items():
+            iface = get_or_create_iface(name)
+            # it comes from the device so we can assume its free of duplicates
```
sebageek: its -> it's

networking_ccloud/ml2/agent/eos/switch.py
```diff
@@ -187,56 +227,107 @@ def _make_config_from_update(self, config):
...
+            if not name.startswith("Port-Channel"):
+                continue
+            iface = get_or_create_iface(name)
+            iface.portchannel_id = int(name[len("Port-Channel"):])
+            iface.members = [p for p in data["ports"] if not p.startswith("Peer")]
+
+        return sorted(iface_map.values(), key=attrgetter('name'))
+
+    def _make_ifaces_config(self, ifaces: Optional[List[agent_msg.IfaceConfig]], operation: Op) -> List[str]:
+        commands = list()
```
sebageek: `list()`

networking_ccloud/ml2/agent/eos/switch.py
```diff
@@ -246,80 +337,38 @@ def get_removable_translations_cmds(iface_name, wanted_translations):
...
-        for name, data in ifaces_vlans.items():
-            if name == "Vxlan1":
-                continue
-            iface = config.get_or_create_iface(name)
-            if data.get("untaggedVlan", 1) != 1:
-                iface.native_vlan = data['untaggedVlan']
-            iface.trunk_vlans = data.get('taggedVlans', [])
+    def _make_config_from_update(self, config: agent_msg.SwitchConfigUpdate) -> List[str]:
+        # build config
+        configmap = dict()
```
sebageek: `{}`

networking_ccloud/ml2/agent/eos/switch.py
```diff
@@ -187,56 +227,107 @@ def _make_config_from_update(self, config):
...
+        for name, data in ifaces_vlan_maps.items():
+            # we're using ingressVlanMappings, but generally we expect them to be the
+            # same as egressVlanMappings
+            for o_vlan, data in data['ingressVlanMappings'].items():
+                result[name].append(agent_msg.VlanTranslation(inside=data['vlanId'], outside=o_vlan))
+        return result
+
+    def get_ifaces_config(self) -> List[agent_msg.IfaceConfig]:
+
+        iface_map = dict()
```
sebageek: `{}`

networking_ccloud/ml2/agent/common/messages.py
swagner-de: I implemented a sort method instead that can be called before making such comparisons.

networking_ccloud/ml2/agent/eos/switch.py
swagner-de: I know its just arguing for basically nothing, but wouldn't the map operation be more streamlined? The attrgetter will be executed at the moment the loop consumes the value while the list comprehension would first materialize the list and then start looping?

swagner-de:
Due to subtle but reoccurring critique of me using `var = list()` and `var = dict()` I will try to restrain from doing that in the future :)

networking_ccloud/ml2/agent/common/messages.py
```diff
@@ -188,6 +197,16 @@ class SwitchConfigUpdate(pydantic.BaseModel):
...
+    def sort(self):
+        # in case we want to compare this or ship it to the user we may benefit from it being sorted
+        # also test cases would fail otherwise
+        if self.vlans:
+            self.vlans.sort()
+        if self.vxlan_maps:
+            self.vxlan_maps.sort()
+        if self.ifaces:
+            self.ifaces.sort()
+
```
sebageek: Shouldn't we sort all members of `SwitchConfigUpdate`? `self.bgp` would need some sorting as well.

networking_ccloud/ml2/agent/eos/switch.py
sebageek: I think as the same work needs to be done in this case it doesn't really matter. If the difference you're talking about is the case that `map()` returns a generator you could also write `(v.vlan for v in device_vlans)`.

sebageek (approved):

PR #56, opened 2022-07-28
Replace eAPI with GNMI for Arista EOS agent

sebageek:


networking_ccloud/ml2/agent/eos/switch.py
```diff
@@ -32,6 +31,121 @@
...
+    class PATHS:
+        VLANS = "network-instances/network-instance[name=default]/vlans"
+        VLAN = "network-instances/network-instance[name=default]/vlans/vlan[vlan-id={vlan}]"
+
+        VXMAP = "interfaces/interface[name=Vxlan1]/arista-exp-eos-vxlan:arista-vxlan/config/vlan-to-vnis"
+        VXMAP_VLAN = ("interfaces/interface[name=Vxlan1]/arista-exp-eos-vxlan:arista-vxlan/config/vlan-to-vnis"
+                      "/vlan-to-vni[vlan={vlan}]")
+
+        EVPN_INSTANCES = "arista/eos/arista-exp-eos-evpn:evpn/evpn-instances"
+        EVPN_INSATNCE = "arista/eos/arista-exp-eos-evpn:evpn/evpn-instances/evpn-instance[name={vlan}]"
```
joker-at-work: variable name contains a typo

networking_ccloud/ml2/agent/eos/switch.py
```diff
@@ -32,6 +31,121 @@
...
+        except Exception as e:
+            LOG.exception("Command %s() failed on %s in %.2fs (retries left: %s): %s (%s)",
+                          method, self._switch_name, time.time() - start_time, retries, e.__class__.__name__, str(e))
+
+            cmd = f"{method}("
+            if args:
+                cmd += ", ".join(map(json.dumps, args))
+            if kwargs:
+                if args:
+                    cmd += " "
```
joker-at-work: Shouldn't this contain a `,`?

networking_ccloud/ml2/agent/eos/switch.py
```diff
@@ -32,6 +31,121 @@
...
+                cmd += ", ".join(f"{k}={json.dumps(v)}" for k, v in kwargs.items())
+            cmd += ")"
+            LOG.debug("Failed command on %s was %s", self._switch_name, cmd)
+            if isinstance(e, AttributeError):
+                # module 'grpc' has no attribute '_channel', sometimes a first-connect problem
+                LOG.info("Reconnecting %s because of %s %s", self._switch_name, e.__class__.__name__, str(e))
```
joker-at-work, line 112: In other LOG lines, you use `%s (%s)` for `e.__class__.__name__, str(e)` - inconsistency detected

networking_ccloud/ml2/agent/eos/switch.py
```diff
@@ -32,6 +31,121 @@
...
+                self.connect()
+            if retries > 0:
+                time.sleep(0.5)
```
joker-at-work, line 115: Want to put this sleep-time into a class-level constant maybe?

networking_ccloud/ml2/agent/eos/switch.py
```diff
@@ -32,6 +31,121 @@
...
+                return self._run_method(method, *args, retries=retries - 1, **kwargs)
+            raise cc_exc.SwitchConnectionError(f"{self._switch_name} {method}() {e.__class__.__name__} {e}")
```
joker-at-work, line 117: Ah, but you also don't use `%s (%s)` here ... hm ... yeah, don't know which is better

networking_ccloud/ml2/agent/eos/switch.py
```diff
@@ -32,6 +31,121 @@
...
+
+    def set(self, *args, retries=3, **kwargs):
+        return self._run_method("set", *args, **kwargs)
+        # if isinstance(ret, grpc._channel._InactiveRpcError):
+        #     if ret.code() in (grpc.StatusCode.UNAVAILABLE, grpc.StatusCode.FAILED_PRECONDITION) \
+        #             and retries > 0:
+        #         time.sleep(0.5)
+        #         return self.set(*args, retries=retries - 1, **kwargs)
+        #     raise ret
+        # return ret
```
joker-at-work: What that do?

networking_ccloud/ml2/agent/eos/switch.py
```diff
@@ -32,6 +31,121 @@
...
+
+
+class EOSSetConfig:
+    def __init__(self):
+        self.delete = []
+        self.replace = []
+        self.update = []
+        self.update_cli = []
+
+    def get_list(self, op):
```
joker-at-work, line 152: Need comment to tell me why only update and replace would be supported as `.delete` and `update_cli` are lists, too ... e.g. "list is not meant as python list but as $something"

joker-at-work (commented):
pre-lunch review. continuing afterwards.

networking_ccloud/ml2/agent/eos/switch.py
```diff
@@ -45,41 +159,13 @@ def get_platform(cls):
...
-            # unpack the response a bit for easier handling
-            # FIXME: will there always be a result? I think so, cause error cases are raised by pyeapi
-            result = result['result']
-            # unpack the response if user only specified a string as cmd
-            if isinstance(cmd, str):
-                result = result[0]
-
-        return result
+        self._api = EOSGNMIClient(f"{self.name} ({self.host})",
+                                  target=(self.host, 6030), username=self.user, password=self._password,
```
joker-at-work, line 175: Did we previously talk about moving that port into a constant?

networking_ccloud/ml2/agent/eos/switch.py
```diff
@@ -32,6 +31,116 @@
 LOG = logging.getLogger(__name__)
 
 
+class EOSGNMIClient:
+
+    class PATHS:
+        VLANS = "network-instances/network-instance[name=default]/vlans"
+        VLAN = "network-instances/network-instance[name=default]/vlans/vlan[vlan-id={vlan}]"
```
joker-at-work, line 38: Should imho be `vlan_id` instead of `vlan`, because you use `vlan` in L42 and ... that might be the name idk, so I'd rather make it clear by naming the variable more explicitly - not relevant up here, but in the usage on L194 you use `vlan=vlan.vlan` and I wasn't sure if we need the id here or the name from the variable we set.

networking_ccloud/ml2/agent/eos/switch.py
```diff
@@ -91,282 +172,342 @@ def get_switch_status(self):
...
+        if operation in (Op.add, Op.replace):
+            if operation == Op.add:
+                # delete all mappings for VNIs we want to repurpose, but are used by a different vlan
+                for curr_map in curr_maps:
+                    for os_map in vxlan_maps:
+                        if os_map.vni == curr_map.vni and os_map.vlan != curr_map.vlan:
+                            LOG.warning("Removing stale vxlan map <vlan %s vni %s> in favor of <vlan %s vni %s> "
+                                        "on switch %s (%s)",
+                                        curr_map.vlan, curr_map.vni, os_map.vlan, os_map.vni, self.name, self.host)
+                            del_map = EOSGNMIClient.PATHS.VXMAP_VLAN.format(vlan=curr_maps.vlan)
```
joker-at-work: This should be `curr_map.vlan` not.

networking_ccloud/ml2/agent/eos/switch.py
```diff
@@ -91,282 +172,342 @@ def get_switch_status(self):
...
+            # delete vlan mapping only if it has the right vni
+            for curr_map in curr_maps:
+                for os_map in vxlan_maps:
+                    if curr_map.vlan == os_map.vlan:
+                        if curr_map.vni == os_map.vni:
+                            config_req.delete.append(EOSGNMIClient.PATHS.VXMAP_VLAN.format(vlan=curr_map.vlan))
+                        else:
+                            LOG.warning("Not deleting vlan %s from switch, as it points to vni %s "
+                                        "(delete requested vni %s)",
+                                        curr_map.vlan, curr_map.vni, os_map.vni)
```
joker-at-work: Missing the switch name and host we log in other logs.

networking_ccloud/ml2/agent/eos/switch.py
```diff
@@ -91,282 +172,342 @@ def get_switch_status(self):
...
+                        break
+                else:
+                    LOG.warning("VLAN %s not found on switch, not deleting it", curr_map.vlan)
```
joker-at-work: Also missing switch, I guess.

networking_ccloud/ml2/agent/eos/switch.py
```diff
@@ -91,282 +172,342 @@ def get_switch_status(self):
...
 
     def get_bgp_vlan_config(self) -> List[agent_msg.BGPVlan]:
-        curr_bgp_vlans = self.send_cmd("show bgp evpn instance")['bgpEvpnInstances']
-        vre = re.compile(r"VLAN (?P<vlan>\d+)")
+        # get potential bgw info per device from
```
joker-at-work: from where?

networking_ccloud/ml2/agent/eos/switch.py
```diff
@@ -91,282 +172,342 @@ def get_switch_status(self):
...
-            bv = agent_msg.BGPVlan(rd=data['rd'], vlan=int(m.group('vlan')),
-                                   rt_imports=data.get('importRts', []), rt_exports=data.get('exportRts', []),
-                                   rd_evpn_domain_all='remoteRd' in data,
-                                   rt_imports_evpn=data.get('importRemoteRts', []),
-                                   rt_exports_evpn=data.get('exportRemoteRts', []))
+            # FIXME: what happens if rd is None?
+            rtdata = entry.get('route-target', {'config': {}})['config']
+            rd = entry['config'].get('route-distinguisher')  # FIXME: this is bad. need to distinguish
+            if not rd:
+                LOG.debug("BGP Vlan %s on switch %s has no rd, skipping it", entry['name'], self.name)
```
joker-at-work, line 276: Switch name was previously followed by its host. Not sure if necessary.

networking_ccloud/ml2/agent/eos/switch.py
```diff
@@ -91,282 +172,342 @@ def get_switch_status(self):
...
+
+    # FIXME: typehint
+    def get_ifaces_config(self, as_dict=False) -> List[agent_msg.IfaceConfig]:
+        ifaces = []
+
+        # get port-channel details
+        pc_details = {}
+        for pc in self.api.get("lacp")['openconfig-lacp:interfaces']['interface']:
+            pc_details[pc['name']] = pc
+        del pc
```
joker-at-work: Can we get a comment why this is necessary and only for the last one?

networking_ccloud/ml2/agent/eos/switch.py
```diff
@@ -91,282 +172,342 @@ def get_switch_status(self):
...
+        # iterate over all ifaces on switch
+        for data in self.api.get(EOSGNMIClient.PATHS.IFACES)['openconfig-interfaces:interface']:
+            iface = agent_msg.IfaceConfig(name=data['name'])
+
+            # port-channel or not?
+            if 'openconfig-if-aggregate:aggregation' in data:
+                data_pc = data['openconfig-if-aggregate:aggregation']
+                data_pc_cfg = data_pc.get('config', {})
+                if not data_pc_cfg.get('arista-intf-augments:mlag') and not data['name'].startswith("Port-Channel"):
+                    LOG.debug("Switch %s LACP iface %s has no mlag id and doesn't start with Port-Channel, skipping it",
```
joker-at-work: Maybe enclose `Port-Channel` in `''`?

networking_ccloud/ml2/agent/eos/switch.py
```diff
@@ -91,282 +172,342 @@ def get_switch_status(self):
...
+                        data['switched-vlan'] = {'config': data_vlan}
+                    iface_cfg = (EOSGNMIClient.PATHS.IFACE_ETH.format(iface=iface_name), data)
+                    config_req.get_list(operation).append(iface_cfg)
+        else:
+            # delete everything (that is requested)
+            def calc_delete_range(all_ifaces, iface_name, iface_cfg):
+                if iface_name not in ifaces:
+                    return []
+                # FIXME: could this list get too big? (for sending the request of unpacked vlans)
+                return list(set(ifaces[iface_name].trunk_vlans) - set(iface_cfg.trunk_vlans))
```
joker-at-work: You're getting the parameter `all_ifaces` here, but are not using it and rely instead onto the outside-scopes function-parameter `ifaces` - is that intended (because it doesn't look like it, as you're passing in `all_ifaces_cfg` down below which you were explicitly retrieving before)?

networking_ccloud/ml2/agent/eos/switch.py
sebageek: Yes, in `pyccloud` we did that. I don't think we need this in the driver. We could put that in once it is actually configurable.

networking_ccloud/ml2/agent/eos/switch.py
sebageek: Why what is necessary? the `del pc`? That was a bad idea (e.g. if there are no lacp interfaces on the box it fails). Or did you mean why we need the data?

joker-at-work: `del pc`, yes.

networking_ccloud/ml2/agent/eos/switch.py
```diff
@@ -32,6 +31,116 @@
...
+            if retries > 0:
+                time.sleep(0.5)
+                return self._run_method(method, *args, retries=retries - 1, **kwargs)
+            raise cc_exc.SwitchConnectionError(f"{self._switch_name} {method}() {e.__class__.__name__} {e}")
+
+    def get(self, *args, unpack=True, single=True, **kwargs):
+        data = self._run_method("get", *args, **kwargs)
+        if data and unpack:
+            data = data['notification']
+            if single:
```
swagner-de, line 120: can we also unpack further if `single=False` like:
```
            if single:
                data = data[0]['update'][0]['val']
            else:
                data = [x['update'][0]['val'] for x in data]
```

networking_ccloud/ml2/agent/eos/switch.py
```diff
@@ -45,41 +154,13 @@ def get_platform(cls):
...
-
-        return result
+        self._api = EOSGNMIClient(f"{self.name} ({self.host})",
+                                  target=(self.host, 6030), username=self.user, password=self._password,
+                                  insecure=False, skip_verify=True)
+        self._api.connect()
 
     def get_switch_status(self):
-        ver = self.send_cmd("show version")
+        ver = self.api.get(path=["cli:/show version"])
```
swagner-de, line 186: I know this one is much simpler, but since we are loving YANG now shouldn't we do it the yang way?
```
data = [x['update'][0]['val'] for x in switch.api.get(path=['system/config/hostname', 'components/component/state/part-no/', 'system/state/boot-time'], single=False)]
data
['qa-de-1-sw4113a-bb271', 'DCS-7050CX3-32S', 1646042091973939968]
```

networking_ccloud/ml2/agent/eos/switch.py
sebageek: Not sure if we can, we'd have to try it with sysdb. With sysdb things are.... somehwat weird, the structure is a bit different and it doesn't play well with single. We could try though.

networking_ccloud/ml2/agent/eos/switch.py
```diff
@@ -32,6 +31,116 @@
 LOG = logging.getLogger(__name__)
 
 
+class EOSGNMIClient:
+
+    class PATHS:
+        VLANS = "network-instances/network-instance[name=default]/vlans"
```
swagner-de, line 37: We could use `network-instances/network-instance[name=default]/vlans[vlan-id=*]/vlan/config` since we are not really interested in all that state data and hence make the API-load on the device a bit smaller.

networking_ccloud/ml2/agent/eos/switch.py
```diff
@@ -32,6 +31,116 @@
...
+                time.sleep(0.5)
+                return self._run_method(method, *args, retries=retries - 1, **kwargs)
+            raise cc_exc.SwitchConnectionError(f"{self._switch_name} {method}() {e.__class__.__name__} {e}")
+
+    def get(self, *args, unpack=True, single=True, **kwargs):
+        data = self._run_method("get", *args, **kwargs)
+        if data and unpack:
+            data = data['notification']
+            if single:
+                data = data[0]['update'][0]['val']
```
swagner-de: can we also unpack further if `single=False` like?
```
            if single:
                data = data[0]['update'][0]['val']
            else:
                data = [x['update'][0]['val'] for x in data]
```

networking_ccloud/ml2/agent/eos/switch.py
```diff
@@ -91,282 +172,399 @@ def get_switch_status(self):
...
+                 for v in swdata]
+        vlans.sort()
+        return vlans
+
+    def _make_vlan_config(self, config_req: EOSSetConfig, vlans: Optional[List[agent_msg.Vlan]], operation: Op) -> None:
+        if vlans is None:
+            return
+
+        if operation in (Op.add, Op.replace):
+            wanted_vlans = []
```
swagner-de, line 209: Not really important, but you could also just do:
```
            vlan_cfg = [(EOSGNMIClient.PATHS.VLAN.format(vlan=vlan.vlan), {'config': {'name': vlan.name, 'vlan-id': vlan.vlan}})
                        for vlan in vlans]
            config_req.get_list(operation).extend(vlan_cfg)
```

networking_ccloud/ml2/agent/eos/switch.py
```diff
@@ -32,6 +31,116 @@
...
+        except Exception as e:
+            raise cc_exc.SwitchConnectionError(f"{self._switch_name} connect() {e.__class__.__name__} {e}")
+
+    def _run_method(self, method, *args, retries=3, **kwargs):
+        try:
+            start_time = time.time()
+            data = getattr(self._gnmi, method)(*args, **kwargs)
+            LOG.debug("Command %s() succeeded in %.2fs", method, time.time() - start_time)
+            return data
+        except Exception as e:
```
swagner-de: Shouldn't we maybe narrow down the exception a bit here?

networking_ccloud/ml2/agent/eos/switch.py
sebageek: Seems to be 14ms faster ion a non-representative benchmark. We'd only need to convert the boottime to uptime-seconds with `time.time() - t / 10 ** 9`

networking_ccloud/ml2/agent/eos/switch.py
```diff
@@ -91,282 +172,399 @@ def get_switch_status(self):
...
+
+    def _make_vxlan_mapping_config(self, config_req: EOSSetConfig, vxlan_maps: Optional[List[agent_msg.VXLANMapping]],
+                                   operation: Op) -> None:
+        if vxlan_maps is None:
+            return
+
+        curr_maps = self.get_vxlan_mappings()
+        if operation in (Op.add, Op.replace):
+            if operation == Op.add:
+                # delete all mappings for VNIs we want to repurpose, but are used by a different vlan
```
swagner-de, line 234: Maybe hint that we cannot use <replace> cause the YANG models key this entity by VLAN, while we key it on vxlan.

networking_ccloud/ml2/agent/eos/switch.py
```diff
@@ -91,282 +172,399 @@ def get_switch_status(self):
...
+                for curr_map in curr_maps:
+                    for os_map in vxlan_maps:
+                        if os_map.vni == curr_map.vni and os_map.vlan != curr_map.vlan:
```
swagner-de, line 237: How about implementing `__eq__` on the class?

networking_ccloud/ml2/agent/eos/switch.py
```diff
@@ -91,282 +172,399 @@ def get_switch_status(self):
...
     def get_bgp_vlan_config(self) -> List[agent_msg.BGPVlan]:
-        curr_bgp_vlans = self.send_cmd("show bgp evpn instance")['bgpEvpnInstances']
-        vre = re.compile(r"VLAN (?P<vlan>\d+)")
+        # FIXME: get potential bgw info per device from sysdb (not properly implemented)
         bgp_vlans = []
-        for vlan_name, data in curr_bgp_vlans.items():
-            m = vre.match(vlan_name)
-            if not m:
-                LOG.warning("Could not match bgp vpn instance name '%s'", vlan_name)
+        curr_bgp_vlans = self.api.get(EOSGNMIClient.PATHS.EVPN_INSTANCES)["arista-exp-eos-evpn:evpn-instance"]
```
swagner-de, line 266: `self.api.get(EOSGNMIClient.PATHS.EVPN_INSTANCES + '/evpn-instance/config', single=False)` gives you just the config. But that would pend a good solution for `api.get`

swagner-de: disregard, there are some nested objects in `EOSGNMIClient.PATHS.EVPN_INSTANCES + '/evpn-instance/` that contain the RT, so going further down 1 path won't work.

networking_ccloud/ml2/agent/eos/switch.py
```diff
@@ -91,282 +172,399 @@ def get_switch_status(self):
...
+            rtdata = entry.get('route-target', {'config': {}})['config']
+            rd = entry['config'].get('route-distinguisher')  # FIXME: this is bad. need to distinguish
+            if not rd:
+                LOG.debug("BGP Vlan %s on switch %s has no rd, skipping it", entry['name'], self.name)
+                continue
+            bv = agent_msg.BGPVlan(rd=rd, vlan=int(entry['name']),
+                                   rt_imports=rtdata.get('import', []), rt_exports=rtdata.get('export', []),
+                                   # eos_native:Sysdb/routing/bgp/macvrf/config/vlan.2323
+                                   rd_evpn_domain_all=False,  # FIXME: get from somewhere
+                                   rt_imports_evpn=[],       # FIXME: get from somewhere
```
swagner-de, line 282: Maybe `entry['route-target']['config']['import']`?

networking_ccloud/ml2/agent/eos/switch.py
```diff
@@ -91,282 +172,399 @@ def get_switch_status(self):
...
+                                   rt_exports_evpn=[])       # FIXME: get from somewhere
+            # FIXME: redistribute learned flag? does not fit in our internal data structure
```
swagner-de, line 284: Feels like in GNMI its a bit more than that? CLI only has `redistribute learned`, but GNMI has `['LEARNED', 'HOST_ROUTE', 'ROUTER_MAC']` and I am not sure if they are implicit but I guess we need to carry them somehow for the diff.

networking_ccloud/ml2/agent/eos/switch.py
```diff
@@ -91,282 +172,399 @@ def get_switch_status(self):
...
             bgp_vlans.append(bv)
         return bgp_vlans
 
     def get_bgp_config(self) -> agent_msg.BGP:
-        bgp_sum = self.send_cmd("show bgp summary")['vrfs']['default']
-        bgp = agent_msg.BGP(asn=bgp_sum['asn'], asn_region=self.drv_conf.global_config.asn_region, vlans=[])
-        bgp.vlans = self.get_bgp_vlan_config()
+        bgp_data = self.api.get(EOSGNMIClient.PATHS.PROTO_BGP)['openconfig-network-instance:bgp']
```
swagner-de: What about this?
```
self.api.get(EOSGNMIClient.PATHS.PROTO_BGP + '/bgp/global/config/as')
4268363793
```

networking_ccloud/ml2/agent/eos/switch.py
```diff
@@ -91,282 +172,399 @@ def get_switch_status(self):
...
+                    "vlans": {
+                        "vlan": [{"vlan-id": bgp_vlan.vlan, "config": {"vlan-id": bgp_vlan.vlan}}],
+                    }
+                }
+
+                cli = [
+                    ("cli:", f"router bgp {bgp.asn}"),
+                    ("cli:", f"vlan {bgp_vlan.vlan}"),
+                ]
+
```
swagner-de, line 316: Should this move under the `if bgp_vlan.rd_evpn_domain_all:` because that indicates that we need the CLI context?

networking_ccloud/ml2/agent/eos/switch.py
```diff
@@ -91,282 +172,399 @@ def get_switch_status(self):
...
+            return
+
+        if operation in (Op.add, Op.replace):
+            evpn_instances = []
+            for bgp_vlan in bgp.vlans:
+                inst = {
+                    "name": str(bgp_vlan.vlan),
+                    "config": {
+                        "name": str(bgp_vlan.vlan),
+                        "redistribute": ["LEARNED"],
```
swagner-de, line 305: This needs `['LEARNED', 'HOST_ROUTE', 'ROUTER_MAC']` for diff consistency.

networking_ccloud/ml2/agent/eos/switch.py
```diff
@@ -91,282 +172,399 @@ def get_switch_status(self):
...
+                        rts["export"] = list(bgp_vlan.rt_exports)
+                    inst["route-target"] = {"config": rts}
+
+                # FIXME: this should be done via model, once we have it
+                # FIXME: what do we do if different route targets exist? they need to go / be removed
+                for rt in bgp_vlan.rt_imports_evpn:
+                    cli.append(("cli:", f"route-target import evpn domain remote {rt}"))
+                for rt in bgp_vlan.rt_exports_evpn:
+                    cli.append(("cli:", f"route-target export evpn domain remote {rt}"))
+                cli.extend([("cli:", "exit"), ("cli:", "exit")])
```
swagner-de, line 339: Do we need that? Coulnd't we just append an `("cli:", "end")` on the very end of the CLI update list or is the CLI context that sensitve to the hierarchy you are currently in?

networking_ccloud/ml2/agent/eos/switch.py
```diff
@@ -91,282 +172,399 @@ def get_switch_status(self):
...
+
+                if bgp_vlan.rd_evpn_domain_all or bgp_vlan.rt_imports_evpn or bgp_vlan.rt_exports_evpn:
+                    # FIXME: I think I checked that this works with replace, but we should make sure
+                    config_req.update_cli.extend(cli)
+                evpn_instances.append(inst)
+            config_req.get_list(operation).append((EOSGNMIClient.PATHS.EVPN_INSTANCES,
```
swagner-de, line 345: Should we maybe split this in a `_make_bgp_vlan_config`-function?

networking_ccloud/ml2/agent/eos/switch.py
sebageek: We rewrap the exception, log that something failed and let the upper class handle it. What would you narrow it down to?

networking_ccloud/ml2/agent/eos/switch.py
sebageek: I mean, that's what it says, doesn't it? We need to delete those mappings and we don't do it for jest. This was also the case with eAPI. I can add a `# This is necessary, as the add fails if the vni is mapped onto another vlan` if you want.

networking_ccloud/ml2/agent/eos/switch.py
sebageek: I actually like the verbose way here. When reading the code I know when it is added and when it's not added. Implementing `__eq__` would hide this imo.

networking_ccloud/ml2/agent/eos/switch.py
sebageek: We don't necessarily `end` it. We might have more cli stuff coming in and as far as I know `end` says "I'm done with configuration"

networking_ccloud/ml2/agent/eos/switch.py
sebageek: `cli` is also used when `rt_exports_evpn` or `rt_imports_evpn` is specified.

networking_ccloud/ml2/agent/eos/switch.py
sebageek: The latter two are set automatically afaik, so we don't need to do that explicitly.

networking_ccloud/ml2/agent/eos/switch.py
sebageek: In both cases it's a vlan number, so I think this is fine.

networking_ccloud/ml2/agent/eos/switch.py
sebageek: As we discussed this somewhat off-github... The method is just a helper to avoid a lot of `if op == ...` and update/replace have a similar structure, while delete looks completely different. If you have a better name for this I'd add it, but from the method I don't feel like it really warrants a comment, as it becomes pretty clear when looking at the rest of the code. I can add one if you feel strongly about it improving readability of the code for someone not that into this driver.

joker-at-work: I don't feel strongly about it, as I'm now more into the code and the usage of this function :D

networking_ccloud/ml2/agent/eos/switch.py
sebageek: Not sure this works properly with a replace, but we can do this in a follow up PR if it works properly

networking_ccloud/ml2/agent/eos/switch.py
sebageek: Does not work properly for some switches as `state/part-no` seems to just not be available. We'll have to investigate this before switching to this, therefore I added a note in the code.

networking_ccloud/ml2/agent/eos/switch.py
sebageek: I think the `vlan-id` on the config-level (not inside) is required if the vlan either exists or does not exist. In one of those cases I think this did not work for me (and am somewhat missing the time to evaluate it at the moment for the potential code gain). Maybe something we can investigate later?

networking_ccloud/ml2/agent/eos/switch.py
sebageek: Yeah, logging names are kind of inconsistent. At somepoint I should overhault the logging and just put this switchname/ip combo in a var everywhere in a consistent fashion. But due to time constraints I'm skipping this now.

networking_ccloud/ml2/agent/eos/switch.py
sebageek: I think they are implicit. They're there by default. Also the cli config only changes if I add `LEARNED`. So I'd go with that. The comment in the code was only referring to adding this to our base config model, as for now we just burn it onto the switch. Not sure it's a priority though.

networking_ccloud/ml2/agent/eos/switch.py
sebageek: I like it!

networking_ccloud/ml2/agent/eos/switch.py
sebageek: TBH I don't see the benefit at this point. I'd leave it, this will require a rewrite anyway once we have the models.

networking_ccloud/ml2/agent/eos/switch.py
sebageek: As discussed off-github this does not contain the evpn stuff that's not yet modeled.

sebageek:
As discussed in Slack - it's time to merge.

PR #65, opened 2022-09-06
Implement metagroup handling for apods

swagner-de:


networking_ccloud/tools/netbox_config_gen.py
```diff
@@ -481,40 +491,49 @@ def make_vpod_metagroup(self, cluster: NbRecord, member: NbRecord) -> Optional[c
...
         if not cname.startswith(cprefix):
             raise ValueError(f'Cluster {cname} must start with {cprefix}')
         cname = cname[len(cprefix):]
         return conf.Hostgroup(binding_hosts=[cname], metagroup=True, members=[member.name])
 
+    def apod_metagroup_handler(self, cluster: NbRecord, member: NbRecord) -> Optional[conf.Hostgroup]:
+        m = self.apod_node_name_re.match(member.name)
+        if not m:
+            raise ValueError(f'{member.name} is not complying with expected node naming for apod clusters')
+        apod_sequence = int(m.group(1))
```
sebageek: If you'd name your catchgroups in the regex you could use `m.group('node')` (or apod) here and it would a bit more clear what's going on :)

networking_ccloud/tools/netbox_config_gen.py
```diff
@@ -481,40 +491,49 @@ def make_vpod_metagroup(self, cluster: NbRecord, member: NbRecord) -> Optional[c
...
+        binding_host_prefix = f'neutron-network-agent-a{apod_sequence:03d}-'
+        # right now 15 network agents are statically created per apod
+        binding_hosts = [f'{binding_host_prefix}{x}' for x in range(0, 15)]
```
sebageek: Maybe make this a constant? Could even be something configurable.

networking_ccloud/tools/netbox_config_gen.py
```diff
@@ -481,40 +491,49 @@ def make_vpod_metagroup(self, cluster: NbRecord, member: NbRecord) -> Optional[c
...
+            if not existing:
+                metagroups[binding_hosts] = metagroup
             else:
-                metagroup.members.append(device.name)
+                existing.members.append(device.name)
 
-        # Check cluster names for duplicates
-        cluster_names = (x.name for x in metagroups.keys())
-        for name, count in Counter(cluster_names).items():
+        # Check cluster names for duplicates5
```
sebageek: `duplicates5` - the new fancy check in town! ;)

networking_ccloud/tools/netbox_config_gen.py
```diff
@@ -481,40 +491,49 @@ def make_vpod_metagroup(self, cluster: NbRecord, member: NbRecord) -> Optional[c
...
-                if not metagroup:
-                    # something failed here, so we ignore
-                    continue
-                metagroups[cluster] = metagroup
+                # something failed here, so we ignore
+                continue
+            # unfortunately the assumption that a netbox cluster corresponds to a metagroup does not hold with
+            # with apods anymore, hence, we call the metagroup handler and then use the binding_host to check
+            # if that metagroup exists
+            binding_hosts = tuple(metagroup.binding_hosts)
```
sebageek: You could also use the property `binding_host_name` of the `Hostgroup` as key if you want:
```
    @property
    def binding_host_name(self):
        """Generate a name for this hostgroup by joining all binding hosts"""
        return ",".join(self.binding_hosts)
```

networking_ccloud/tools/netbox_config_gen.py
```diff
@@ -458,63 +502,42 @@ def get_interconnect_hostgroups(self, nb_switches: List[NbRecord]) -> List[conf.
...
             else:
-                metagroup.members.append(device.name)
+                existing.members.append(device.name)
 
         # Check cluster names for duplicates
-        cluster_names = (x.name for x in metagroups.keys())
-        for name, count in Counter(cluster_names).items():
+        for name, count in Counter(metagroups.keys()).items():
             if count > 1:
                 raise ConfigException(f'Cluster {name} appeared {count} times')
```
sebageek: I don't think this check will ever work, as you are checking if there are multiple entries with the same key in a dictionary. If you'd try that, you'd just override the old entry.

swagner-de: Indeed, it did not do anything that made sense earlier and it does not make sense now.

sebageek (approved):

PR #78, opened 2022-09-21
Add common functionality into the backdoor shell

joker-at-work:
We define a couple of general-purpose functions for the debug shell and include them into the shell's locals for easy use.

### DIFF
--- a/networking_ccloud/ml2/agent/common/backdoor.py
+++ b/networking_ccloud/ml2/agent/common/backdoor.py
@@ -1,0 +1,90 @@
+# Copyright 2022 SAP SE
+#
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#    under the License.
+from collections import Counter
+import gc
+import sys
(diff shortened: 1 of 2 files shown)


networking_ccloud/ml2/agent/common/backdoor.py
```diff
@@ -0,0 +1,90 @@
...
+def _print_semaphores():
+    """Print all Semaphore objects used by oslo_concurrency.lockutils and their waiter count"""
+    # local import as we don't want to keep that local variable in global scope
+    from oslo_concurrency.lockutils import _semaphores
+
+    print('\n'.join(sorted([f"{name} - {len(s._Semaphore__cond._Condition__waiters)}"
+                            for name, s in _semaphores._semaphores.items()])))
+
+
+def _time_it(fn):
```
sebageek: Maybe allow `*args, **kwargs` and add that to the `fn()` call? I know you mainly use it with a lambda call, but I think this would fit in nicely, as it also mimics how python testing does this in some places. Same would go for `_profile_it()`.

networking_ccloud/ml2/agent/common/backdoor.py
```diff
@@ -0,0 +1,90 @@
...
+
+    pr = cProfile.Profile()
+    pr.runcall(fn)
+    pr.create_stats()
+    ps = pstats.Stats(pr)
+
+    if return_stats:
+        return ps
+
+    ps.sort_stats('tottime').print_stats(30)
```
sebageek, line 74: Should this be conditional on `return_stats`? Should we make this more configurable? Or is `tottime` that what we always want and `30` always the amount?

joker-at-work: This is just a simple default to get a quick overview of things. If you want to know something else or do something more fancy, set `return_stats` and do something with it on your own ;)

sebageek (approved):

PR #96, opened 2022-09-29
Save config switch periodic task

sebageek:
Two commits are extra here, as my PR depends on  #94. That PR will be merged before this one.

### DIFF
--- a/networking_ccloud/common/config/config_oslo.py
+++ b/networking_ccloud/common/config/config_oslo.py
@@ -23,8 +23,14 @@ cc_fabric_opts = [
                help="Path to yaml config file"),
 ]
 
+cc_fabric_agent_opts = [
+    cfg.IntOpt('persist_config_loop_interval', default=15 * 60,
+               help="Interval between config persists (use -1 to disable)"),
+]
+
 # make sure we have ml2 vlan opts available before this option is parsed
 # as we need to access ml2_type_vlan.network_vlan_ranges in our validation
 register_ml2_drivers_vlan_opts()
 
 cfg.CONF.register_opts(cc_fabric_opts, "ml2_cc_fabric")
+cfg.CONF.register_opts(cc_fabric_agent_opts, "ml2_cc_fabric_agent")
--- a/networking_ccloud/ml2/agent/common/agent.py
+++ b/networking_ccloud/ml2/agent/common/agent.py
@@ -14,12 +14,14 @@
 
 from operator import attrgetter
 import sys
+import time
 
 from neutron.common import config as common_config
 from neutron.conf.agent.common import register_agent_state_opts_helper
 from neutron import manager
 from oslo_config import cfg
 from oslo_log import log as logging
+from oslo_service import periodic_task
 
 from networking_ccloud.common.config import get_driver_config
 from networking_ccloud.common import constants as cc_const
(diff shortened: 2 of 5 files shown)


networking_ccloud/common/config/config_oslo.py
```diff
@@ -23,8 +23,16 @@
                help="Path to yaml config file"),
 ]
 
+cc_fabric_agent_opts = [
+    cfg.BoolOpt('persist_config_loop_enabled', default=True,
+                help="Enable/Disable the loop that tells the switch to persists its config"),
+    cfg.IntOpt('persist_config_loop_interval', default=15 * 60,
+               help="Interval between config persists, if loop is enabled"),
+]
```
joker-at-work: In Nova, this is usually not handled by another variable, but by setting `persist_config_loop_interval` to a negative number, e.g. [here](https://github.com/sapcc/nova/blob/ba9c8e612e0d0e86010b076077b255e409cdc493/nova/conf/compute.py#L770-L773)

networking_ccloud/ml2/agent/common/agent.py
```diff
@@ -177,6 +179,21 @@ def apply_config_update(self, context, config):
...
+        if not cfg.CONF.ml2_cc_fabric_agent.persist_config_loop_enabled:
+            return
+
+        start_time = time.time()
+        LOG.info("Persisting config on all switches")
+        futures = []
+        for switch in sorted(self._switches, key=attrgetter('name')):
+            futures.append(switch.persist_config())
+        for future in futures:
+            future.result()
```
joker-at-work, line 191: This might raise an exception. Not sure if you don't want to wait for the rest? They will run nonetheless.

networking_ccloud/ml2/agent/common/switch.py
```diff
@@ -111,3 +112,15 @@ def apply_config_update(self, config):
...
         raise NotImplementedError
+
+    @run_in_executor('write')
+    def persist_config(self):
+        start_time = time.time()
+        try:
+            self._persist_config()
+            LOG.debug("Switch config of %s saved in %.2f", self, time.time() - start_time)
+        except Exception as e:
+            LOG.error("Saving switch config of %s failed in %.2f: %s", self, time.time() - start_time, e)
```
joker-at-work, line 123: You sure you don't want `LOG.exception`? `KeyError`s area otherwise quite hard to debug ...
```
>>> try:
...   {}['asdf']
... except Exception as e:
...   print(str(e))
... 
'asdf'
```

joker-at-work (commented):
commit-message:
> For EOS it will simply
> to a "copy running-config startup-config".

`to` -> `do`

networking_ccloud/common/config/config_oslo.py
sebageek: I thought about doing it this way, but deemed it "user unfriendly". But if this is the OpenStack default, then I happily oblige :)

networking_ccloud/ml2/agent/common/switch.py
sebageek: I only expect there to be GNMI errors and I think these don't need a backtrace logged. We also already log the backtrace for GNMI commands. So until the switch config saving get's any more complex I'd leave it this way.

networking_ccloud/ml2/agent/common/agent.py
sebageek: As we already catch everything in the current implementation of `persist_config()` I think this is fine.

joker-at-work (approved):

PR #97, opened 2022-09-30
Let a full sync flush the ThreadPool

joker-at-work:
Whenever we schedule a full-sync, we want - nearly - all other tasks to be flushed out of the write ThreadPool for a switch, because the full-sync will do the changes other tasks would do anyways.

To implement this, we use a variable guarded by a lock. This variable is either set to None if no full-sync is currently waiting to run or to a Future object for the scheduled full-sync.

Before scheduling a new task, a thread checks the guarded variable to see if a full-sync is scheduled already and returns the full-sync Future instead if that's the case.

Already scheduled tasks check once they start running, if there's a full-sync waiting to run by checking the guarded variable. If they find a scheduled full-sync, they raise FullSyncScheduled which contains the full-sync Future in the "future" attribute. Callers can then decide if they return immediately or wait for the full-sync to finish - but they have to catch that exception.

There are 2 ways a full-sync might be scheduled: the normal way either via RPC or periodic task OR by a task getting scheduled and finding the queue in bad shape. Whenever we schedule a full-sync, the function "run_full_sync" takes the lock of the guarded variable. If the function finds the guarded variable already set after taking the lock, it returns the Future, because there's no need to schedule another full-sync. If the variable isn't set, it submits the full-sync function to the ThreadPool (i.e. it schedules the full-sync) and sets the variable to the returned Future.

To make sure we don't lose any updates, a starting full-sync will unset the guarded variable before doing any actual work. This makes sure that updates of tasks getting submitted in the middle of a full-sync will not be missed until next full-sync.

### DIFF
--- a/networking_ccloud/ml2/agent/common/switch.py
+++ b/networking_ccloud/ml2/agent/common/switch.py
@@ -14,9 +14,11 @@
 
 import abc
 from functools import wraps
+import threading
 import time
 
 from futurist import ThreadPoolExecutor
+from neutron_lib.context import get_admin_context
 from oslo_concurrency import lockutils
 from oslo_context import context
 from oslo_log import log as logging
@@ -26,20 +28,68 @@ from networking_ccloud.ml2.driver_rpc_api import CCFabricDriverRPCClient
 LOG = logging.getLogger(__name__)
 
 
-def run_in_executor(type_):
+class FullSyncScheduled(Exception):
+    """Raised if a scheduled write-attempt was voluntarily interrupted because there's a full-sync in the queue
+
+    Callers catching this exception should wait for the attached future
+    attribute instead to know when their task finishes.
+    """
+    def __init__(self, *, future):
+        super().__init__(self)
+        self.future = future
+
+
+def run_in_executor(type_, replacable_by_full_sync=False):
     """Decorator to run a method in the ThreadPoolExecutor of the class
 
     The wrapped method thus returns a futurist.Future now instead of its actual
     result.
+
+    If replacable_by_full_sync is set, the decorated method will not run if
+    there's a full-sync scheduled. Instead, it will raise FullSyncScheduled.
+    Additionally, the decorated method will also not get scheduled, if there's
+    already a full-sync scheduled. In that case, the wrapped method returns the
+    scheduled full-sync's futurist.Future instead of its own. Callers have to
+    be able to handle that.
     """
+    if type_ != 'write' and replacable_by_full_sync:
+        raise ValueError(f"@run_in_executor() called with replacable_by_full_sync=True on {type_} executor - "
+                         "only 'write' is supported.")
+
     def decorator(fn):
 
         @wraps(fn)
         def wrapped(self, *args, **kwargs):
+            executor = getattr(self, f'_{type_}_executor')
+
             current_context = context.get_current()
 
+            if replacable_by_full_sync:
+                # NOTE(jkulik): This is a very rough algorithm as a starting
+                # point and up for discussion. It assumes that a full-sync is
+                # a better/faster solution than running all the tasks in the
+                # queue. It only holds true for tasks that are
+                # replacable_by_full_sync, but the assumption is that these are
+                # most of the tasks. Average runtime is skewed by having the
+                # full-sync run in the same queue.
+                avg_runtime = executor.statistics.average_runtime if executor.statistics.executed else 0
+                qsize = executor._work_queue.qsize()
+                full_sync_threshold = 120
+                if qsize * avg_runtime > full_sync_threshold:
+                    LOG.warning("Scheduling full-sync on %s for queue_size %s and avg runtime %s as the threshold "
+                                "of %s is reached", self, qsize, avg_runtime, full_sync_threshold)
+                    return self.run_full_sync(current_context.elevated() if current_context else get_admin_context())
+
+                # if there's currently a full-sync scheduled and not started,
+                # yet, we do not schedule our changes but rely on the full-sync
+                # instead
+                full_sync_future = self.get_full_sync_future()
+                if full_sync_future is not None:
+                    LOG.debug("Found scheduled full-sync on %s. Returning full-sync's Future instead", self)
+                    return full_sync_future
+
             @wraps(fn)
-            def context_preserving_fn(*args, **kwargs):
+            def extended_fn(self, *args, **kwargs):
                 if current_context:
                     current_context.update_store()
                 else:
@@ -49,10 +99,18 @@ def run_in_executor(type_):
                         delattr(context._request_store, 'context')
                     except AttributeError:
                         pass
-                return fn(*args, **kwargs)
 
-            executor = getattr(self, f'_{type_}_executor')
-            return executor.submit(context_preserving_fn, self, *args, **kwargs)
+                if replacable_by_full_sync:
+                    # short-circuit this function if there's a full-sync in the
+                    # pipeline. our changes will be included there anyways
+                    full_sync_future = self.get_full_sync_future()
+                    if full_sync_future is not None:
+                        LOG.debug("Returning early for scheduled full-sync on %s", self)
+                        raise FullSyncScheduled(future=full_sync_future)
+
+                return fn(self, *args, **kwargs)
+
+            return executor.submit(extended_fn, self, *args, **kwargs)
 
         return wrapped
 
@@ -73,6 +131,8 @@ class SwitchBase(abc.ABC):
         self._api = None
         self._read_executor = ThreadPoolExecutor(max_workers=5)
         self._write_executor = ThreadPoolExecutor(max_workers=1)
+        self._full_sync_future_lock = threading.Lock()
+        self._full_sync_future = None
 
         self._rpc_client = CCFabricDriverRPCClient()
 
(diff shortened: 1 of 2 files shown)


networking_ccloud/ml2/agent/common/switch.py
```diff
@@ -26,20 +27,59 @@
 LOG = logging.getLogger(__name__)
 
 
-def run_in_executor(type_):
+class FullSyncScheduled(Exception):
+    """Raised if a scheduled write-attempt was voluntarily interrupted because there's a full-sync in the queue
+
+    Callers catching this exception should wait for the attached future
+    attribute instead to know when their task finished.
```
sebageek: Either "to know when their task is finished" or "when their task finishes". I think "finished" can't be past tense here.

networking_ccloud/ml2/agent/common/switch.py
```diff
@@ -26,20 +27,59 @@
...
+    """
+    def __init__(self, *args, future, **kwargs):
+        super().__init__(self, *args, **kwargs)
```
sebageek: Do we really want to tansport custom additional messages in this exception? I think leave the kwargs out here and maybe - just maybe - add a custom message to it explaining what this is if this ever would not get caught and end up in our logs.

networking_ccloud/ml2/agent/common/switch.py
```diff
@@ -26,20 +27,59 @@
...
+def run_in_executor(type_, replacable_by_full_sync=False):
     """Decorator to run a method in the ThreadPoolExecutor of the class
 
     The wrapped method thus returns a futurist.Future now instead of its actual
     result.
+
+    If replacable_by_full_sync is set, the decorated method will not run if
+    there's a full-sync scheduled. Instead, it will raise FullSyncScheduled.
+    Additionally, the decorated method will also not get scheduled, if there's
+    full-sync scheduled already. In that case, the wrapped method returns the
```
sebageek: "If there's _a_ full-sync scheduled". I also don't know if the already needs to be earlier in that sentence.

networking_ccloud/ml2/agent/common/switch.py
```diff
@@ -26,20 +27,59 @@
...
+    futurist.Future of the scheduled full-sync instead of of its own. Callers
```
sebageek: of of

networking_ccloud/ml2/agent/common/switch.py
```diff
@@ -26,20 +27,59 @@
...
+    have to be able to handle that.
     """
     def decorator(fn):
 
         @wraps(fn)
         def wrapped(self, *args, **kwargs):
+            executor = getattr(self, f'_{type_}_executor')
+
+            if replacable_by_full_sync and type_ == 'write':
+                # NOTE(jkulik): This is a very rough algorithm as starting
```
sebageek: "as _a_ starting point"

networking_ccloud/ml2/agent/common/switch.py
```diff
@@ -26,20 +27,59 @@
...
+                # point and up for discussion that assumes that a full-sync is
```
sebageek: "and _it is_ up for discussion, that..."

networking_ccloud/ml2/agent/common/switch.py
```diff
@@ -26,20 +27,59 @@
...
+                # a better/faster solution than running all the tasks in the
+                # queue. It only holds true for tasks that are
+                # replacable_by_full_sync, but the assumption is that these are
+                # most of the tasks. Average runtime is skewed by having the
+                # full-sync run in the same queue.
+                avg_runtime = executor.statistics.average_runtime if executor.statistics.executed else 0
+                if executor._work_queue.qsize() * avg_runtime > 120:
+                    return self.run_full_sync()
```
sebageek: I think this should produce a log message. I'd go with warning, info would be acceptable. Log message should also print out `qsize()` and `avg_runtime`, maybe also the result of it + what it does ("scheduling full sync because...).

networking_ccloud/ml2/agent/common/switch.py
```diff
@@ -49,10 +89,17 @@ def context_preserving_fn(*args, **kwargs):
                         delattr(context._request_store, 'context')
                     except AttributeError:
                         pass
-                return fn(*args, **kwargs)
 
-            executor = getattr(self, f'_{type_}_executor')
-            return executor.submit(context_preserving_fn, self, *args, **kwargs)
+                if replacable_by_full_sync and type_ == 'write':
```
sebageek: If the dev using this already says `replacable_by_full_sync`, do we really need to ensure this is a write? I agree with "all replaceable calls should be writes", but do we need to double check or shall we just assume that the dev knows what they're doing when they mark a read-call as replacable by full sync?

Now that I'm thinking about it, maybe it makes sense, as we explicitly only want to do thinks in the write queue? Hmmm. Not quite sure. Thoughts?

networking_ccloud/ml2/agent/common/switch.py
```diff
@@ -129,10 +182,29 @@ def persist_config(self):
...
+
+        @run_in_executor('write')
+        def _in_thread_full_sync(self, context):
+            # once the full sync starts, we remove access to the future,
+            # because new state coming in might not be seen by a running
+            # full-sync anymore and thus be missed
+            with self._full_sync_future_lock:
+                self._full_sync_future = None
+
+            return self._run_full_sync(context)
```
sebageek, line 210: With this we reset the full sync future before actually running the sync, so we miss out on catching everyone that joins the party™ while the full sync runs. Wouldn't it make more sense to reset the full sync future once we have the result (or an exception) of `_run_full_sync()`, but before returning (i.e. concluding the future)? Could this be a case for `try`/`finally`?

sebageek: Ah, I just now have "been read" the commit message. We want to make sure we don't loose any updates and that could happen, because updates _after_ fetching the config still need to be applied. Totally makes sense!

networking_ccloud/ml2/agent/common/switch.py
joker-at-work: Actually ... this could be fine ;) "instead of the futurist.Future of its own task" with "the futurist.Future" left out, because it's implied. rephrased a little.

networking_ccloud/ml2/agent/common/switch.py
joker-at-work: "it is" is implied from the previous part of the sentence. I don't feel like adding "it is" helps. I'll rephrase that part with the "that" though.

networking_ccloud/ml2/agent/common/switch.py
joker-at-work: I've opted for raising a `ValueError` in the decorator-creation if `type_ != 'write' and replacable_by_full_sync`. I think that should align with your remarks.

So, yes, I think we should make the interface/restrictions clear in the function itself and not only rely on the dev knowing what they're doing. Having that exception raised should help the dev in seeing that they don't ;)

networking_ccloud/ml2/agent/common/switch.py
```diff
@@ -26,20 +28,67 @@
...
     def decorator(fn):
 
         @wraps(fn)
         def wrapped(self, *args, **kwargs):
+            executor = getattr(self, f'_{type_}_executor')
+
             current_context = context.get_current()
 
+            if replacable_by_full_sync:
+                # NOTE(jkulik): This is a very rough algorithm as starting a
```
sebageek: ..."as a starting point"

networking_ccloud/ml2/agent/common/switch.py
```diff
@@ -26,20 +28,67 @@
...
+                # queue. It only holds true for tasks that are
+                # replacable_by_full_sync, but the assumption is that these are
+                # most of the tasks. Average runtime is skewed by having the
+                # full-sync run in the same queue.
+                avg_runtime = executor.statistics.average_runtime if executor.statistics.executed else 0
+                qsize = executor._work_queue.qsize()
+                full_sync_threshold = 120
+                if qsize * avg_runtime > full_sync_threshold:
+                    LOG.warning('Scheduling full-sync for queue_size %s and avg runtime %s as the threshold of %s is '
+                                'reached', qsize, avg_runtime, full_sync_threshold)
```
sebageek: Could you also ad the switch name here, so we know which switch exceeded the size? Btw, printing `self` should do the needful there, afair it has a proper `__str__()` that prints out name and ip.

networking_ccloud/ml2/agent/common/switch.py
```diff
@@ -49,10 +98,17 @@ def context_preserving_fn(*args, **kwargs):
...
-                return fn(*args, **kwargs)
 
-            executor = getattr(self, f'_{type_}_executor')
-            return executor.submit(context_preserving_fn, self, *args, **kwargs)
+                if replacable_by_full_sync:
+                    # short-circuit this function if there's a full-sync in the
+                    # pipeline. our changes will be included there anyways
+                    full_sync_future = self.get_full_sync_future()
+                    if full_sync_future is not None:
+                        raise FullSyncScheduled(full_sync_future)
```
sebageek: Needs to be a kwarg

sebageek (approved):

sebageek:
Tests are still missing for this, but for now we want to test the functionality in qa, so tests will have to be added later.

PR #102, opened 2022-10-10
Pass config context for vlan range override to driver config

swagner-de:
The field in the netbox config context for vlan range override nis now utilized and passed on in the driver config. Also slightly adjusting the docs to follow the formatting.

### DIFF
--- a/networking_ccloud/tools/netbox_config_gen.py
+++ b/networking_ccloud/tools/netbox_config_gen.py
@@ -363,6 +363,7 @@ class ConfigGenerator:
         attributes = dict()
         pod_roles = set()
         azs = set()
+        vlan_ranges = set()
 
         for device in devices:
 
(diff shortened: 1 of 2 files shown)


networking_ccloud/tools/netbox_config_gen.py
```diff
@@ -375,6 +376,16 @@ def get_switchgroup_attributes(cls, devices: List[NbRecord]) -> Dict[str, str]:
                 raise ConfigException(f'{device.name} must have a site')
             azs.add(device.site.slug)
 
+            # segment ranges
+            if device.config_context:
+                vlan_range = device.config_context.get('cc', {}) \
+                                                  .get('net', {}) \
+                                                  .get('evpn', {}) \
+                                                  .get('tenant-vlan-range', {})
```
sebageek: I think the final default parameter should not be a dict, I'd leave it unset.

networking_ccloud/tools/netbox_config_gen.py
```diff
@@ -375,6 +376,16 @@ def get_switchgroup_attributes(cls, devices: List[NbRecord]) -> Dict[str, str]:
...
+                if vlan_range:
+                    vlan_range = tuple(sorted(vlan_range))
+                    vlan_ranges.add(vlan_range)
```
sebageek, line 395: how do the vlan ranges look? Do we need to do any format changes at this point or shall we just let that be done by the pydanic config checks?

networking_ccloud/tools/netbox_config_gen.py
```diff
@@ -383,7 +394,10 @@ def get_switchgroup_attributes(cls, devices: List[NbRecord]) -> Dict[str, str]:
                                                          ', '.join((x.name for x in devices)), 'pod_roles')
         attributes['az'] = cls._ensure_single_item(azs, 'Switchgroup',
                                                    ', '.join((x.name for x in devices)), 'azs')
-
+        if vlan_ranges:
+            attributes['vlan_ranges'] = list(cls._ensure_single_item(vlan_ranges, 'Switchgroup',
+                                                                     ', ' .join((x.name for x in devices)),
```
sebageek: No need to double-bag.

networking_ccloud/tools/netbox_config_gen.py
swagner-de: I updated the docs to reflect that ranges are using `:` as the range indicator. I would not do any sanitizing in the config generator.

sebageek: Aye, I'm not against that. But if the config_context is broken (e.g. not a list) this would lead to an exception and we probably don't know where it comes from (no "switch x has broken context"), but that'll only happen if people put in garbage into the context.

sebageek (approved):

PR #105, opened 2022-10-13
Modelling and config generation for loadbalancers

swagner-de:


networking_ccloud/tools/netbox_model.py
```diff
@@ -475,6 +475,30 @@ def model_neutron_routers(self, limit: Optional[Set[int]] = None):
...
+                continue
+            if not self.ensure_single_attribute(('site', 'slug'), gr):
+                print(f'Np{np} switches {gr} have different sites, skipping')
+                continue
+            lags = self.find_and_bundle_mlag_ports(np, gr, 'loadbalancer')
+            for lag, members in lags:
+                for member in members:
+                    # We only care about portchannel1 at the moment cause that's the one OS binds ports to
+                    ignore_tag = list(ConfigGenerator.ignore_tags)[0]
+                    if member.connected_endpoint.lag.name != 'portchannel1':
```
sebageek: In cases like this I'd go for the version with less indentation for the later code, i.e.
```python
if member == 'port-channel1:
    continue

if isinstance(lag, dict):
    ...
...
```

networking_ccloud/tools/netbox_config_gen.py
```diff
@@ -129,7 +159,8 @@ def noop_metagroup_handler(self, cluster: NbRecord, member: NbRecord) -> Optiona
         'filer': filer_metagroup_handler,
         'cc-vsphere-apod-mgmt': noop_metagroup_handler,
         'cc-vsphere-apod-pool': noop_metagroup_handler,
-        'cc-k8s-controlplane-swift': noop_metagroup_handler
+        'cc-k8s-controlplane-swift': noop_metagroup_handler,
+        'cc-f5-vcmp': f5_loadbalancer_metagroup_handler
```
sebageek: As we had this a couple times, adding a `,` after the last element would only show exactly that line each time we add a hostgroup handler here ;)

sebageek (approved):

PR #110, opened 2022-10-25
Config-Gen/Modeller Ignore non LAGs, ignore foreign LAG ids

swagner-de:


networking_ccloud/tools/netbox_config_gen.py
```diff
@@ -58,6 +59,7 @@ class ConfigGenerator:
     switch_name_re = re.compile(r"^(?P<region>\w{2}-\w{2}-\d)-sw(?P<switchgroup_id>(?P<az>\d)"
                                 r"(?P<pod>\d)(?P<switchgroup>\d{2}))"
                                 r"(?P<leaf>[ab])(?:-(?P<role>[a-z]+(?P<seq_no>[0-9]+)?))$")
+    lag_id_re = re.compile(r"^\S+?(?P<lag_id>\d+)$")
```
sebageek, line 62: `\S` also matches numbers and `\S+` is greedy. I think you explicitly need to exclude them or use something like `[a-z-A-Z/-]`.

```
In [12]: re.match(r"(\S+)(\d+)", "Port-Channel333").groups()
Out[12]: ('Port-Channel33', '3')
```

networking_ccloud/tools/netbox_config_gen.py
```diff
@@ -528,6 +530,18 @@ def get_connected_devices(self, switches: List[NbRecord]) -> Tuple[Set[NbRecord]
                 # Those that we find unbundled are probably not modelled correctly, so we exclude them.
                 if not iface.lag:
                     continue
+                m = self.lag_id_re.match(iface.lag.name)
+                if not m:
+                    print(f'Cannot parse LAG id on interface {switch.name}/{iface.name}, '
+                          f'bundled in {iface.lag.name}')
+                lag_id = int(m.group('lag_id'))
+
+                if not RESPONSIBLE_LAG_RANGES[0] < lag_id < RESPONSIBLE_LAG_RANGES[1]:
```
sebageek: So the lag ranges are exclusive on both ends? I'd expect them to be inclusive on both ends, as with most (if not all) our other ranges in this piece of software.

networking_ccloud/tools/netbox_config_gen.py
```diff
@@ -528,6 +530,18 @@ def get_connected_devices(self, switches: List[NbRecord]) -> Tuple[Set[NbRecord]
...
+                    if self.verbose:
+                        print(f'LAG id {lag_id} on {switch.name}/{iface.lag.name} is oustide of driver '
+                              f'responsible bounds [{RESPONSIBLE_LAG_RANGES[0]}, {RESPONSIBLE_LAG_RANGES[1]}]')
```
sebageek, line 542: Mathematically this would also hint for an inclusive range, as `[a, b]` is generally used to denote that in math. `(a, b)` would be exkclusive.

https://en.wikipedia.org/wiki/Interval_(mathematics)#Including_or_excluding_endpoints

networking_ccloud/tools/netbox_config_gen.py
swagner-de: Yes, but I am already using the lazy quantifier which matches just fine
```
>>> import re
>>> lag_id_re = re.compile(r"^\S+?(?P<lag_id>\d+)$")
>>> lag_id_re.match("Port-Channel333").groups()
('333',)
```

networking_ccloud/tools/netbox_config_gen.py
swagner-de: They indeed are exclusive. While technically our range is 100-199, inclusive, we will never allow a LAG of ID 100 as per our LAG ID derivation that LAG would need its first interface on interface index 0, which does not exists. Hence the range is exclusive.

networking_ccloud/tools/netbox_config_gen.py
swagner-de: Fair point.

networking_ccloud/tools/netbox_config_gen.py
sebageek: What do you think about making them inclusive and changing the range to `(101, 199)`?

sebageek (approved):

PR #111, opened 2022-10-25
[config-gen] Import address-scope from external YAML file into VRF conf

swagner-de:
We have to place in netbox to maintain address scopes in the moment. We maintain it in our config repo in a YAML file, hence we reference that file here and enrich the VRF config with the address scopes from that YAML.

### DIFF
--- a/doc/source/configuration/config-gen.rst
+++ b/doc/source/configuration/config-gen.rst
@@ -17,12 +17,15 @@ It does support nesting the configuration under certain YAML keys if the config 
 
 Vault references can also be injected instead of plain-text passwords using the :code:`--vault-ref` parameter.
 
+In our environment, we also need a mapping from OpenStack address-scopes to externally routable VRFs. We can supply that mapping using the :code:`--address-scope-vrf-map` parameter.
+Multiple mappings can be supplied. The Config Generator will check the YAML file for any top level key that is :code:`address_scopes`, :code:`local_address_scopes` or :code:`global_address_scopes` and will then try to parse a VRF and an address scope name out of it. For this to work, the list items must have a :code:`name` and :code:`vrf` key.
+
 Example
 --------
 
 ::
 
-   cc-netbox-config-gen -r $OS_REGION_NAME -u MYAPIUSER -V $OS_REGION_NAME/my/vault/path -w cc_fabric/driver_config -o $PATH_TO_HELM_VALUES/$OS_REGION_NAME/values/neutron-cc-fabric.yaml
+   cc-netbox-config-gen -r $OS_REGION_NAME -u MYAPIUSER -V $OS_REGION_NAME/my/vault/path -a $PYCCLOUD_SECRETS_REPO_PATH/global/values/neutron-networking-global.yaml -w cc_fabric/driver_config -o $PATH_TO_HELM_VALUES/$OS_REGION_NAME/values/neutron-cc-fabric.yaml
 
 
 
--- a/networking_ccloud/tools/netbox_config_gen.py
+++ b/networking_ccloud/tools/netbox_config_gen.py
@@ -17,6 +17,7 @@ import ipaddress
 from itertools import chain, groupby
 import logging
 from operator import attrgetter, itemgetter
+from pathlib import Path
 import re
 import urllib3
 from typing import Any, Dict, Generator, Iterable, List, Optional, Set, Tuple, Union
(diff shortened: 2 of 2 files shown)


networking_ccloud/tools/netbox_config_gen.py
```diff
@@ -781,6 +810,9 @@ def main():
 
     parser.add_argument('-w', '--wrap-in', help='Keys under which the generated config should be nested under. '
                                                 'Format should be: <key1>/<key2>...')
+    parser.add_argument("-a", "--address-scope-vrf-map", type=Path, nargs="+",
+                        help="Path to file containing a mapping of address scope names to VRFs. If this is omitted, "
+                             "no mapping will be generated.")
```
sebageek, line 816: Why do we need to allow multiple files here?

networking_ccloud/tools/netbox_config_gen.py
```diff
@@ -213,13 +214,39 @@ def get_azs(self) -> List[conf.AvailabilityZone]:
...
 
-    def get_cloud_vrfs(self) -> List[conf.VRF]:
+    def get_address_scopes(self, address_scope_vrf_maps_path: List[Path]) -> Dict[str, List[str]]:
+        result = defaultdict(list)
+        for p in address_scope_vrf_maps_path:
+            with p.open('r') as f:
+                asvm = yaml.safe_load(f)
+                for k, values in asvm.items():
+                    if not isinstance(k, str) or not k.endswith('address_scopes'):
+                        continue
```
sebageek: Why do the search for `.*address_scopes`? We have an expected format for this file. I'd check for the key `global_address_scopes` and if it's not there raise an error.

networking_ccloud/tools/netbox_config_gen.py
```diff
@@ -213,13 +214,39 @@ def get_azs(self) -> List[conf.AvailabilityZone]:
...
+                        vrf = item.get('vrf')
+                        if not name:
+                            print(f'{p.as_posix()}, key {k}, item {i}, needs a "name" to be considered '
+                                  'a valid address-scope mapping')
+                            continue
+                        if not vrf:
+                            print(f'{p.as_posix()}, key {k}, item {i}, needs a "vrf" to be considered '
+                                  'a valid address-scope mapping')
+                            continue
+                        result[vrf.lower()].append(name)
```
sebageek, line 237: I tried building a shorter solution for this but it did end up not really being shorter. But for fun and giggles:
```python
for i, item in enumerate(values):
    valid = False
    for key in 'name', 'vrf':
        if not item.get(key):
            print(f'{p.as_posix()}, key {k}, item {i}, needs a "{key}" to be considered '
                  'a valid address-scope mapping')
            break
    else:
        valid = True
    if not valid:
        continue

    result[item['vrf'].lower()].append(item['name'])
```
So... let's keep yours I guess ;)

networking_ccloud/tools/netbox_config_gen.py
swagner-de: Our helm charts have the option to use the global address scopes and local address scopes as well. We do not use it, I hope we never will, but I wanted to implement the feature here too.

networking_ccloud/tools/netbox_config_gen.py
swagner-de: Same as in above comment. helm-charts allow 2 places to define address scopes, once with `global_address_scopes` once with `local_address_scopes`. To also support if somebody just wants a YAML file with the key `address_scopes` I implemented it that way.

networking_ccloud/tools/netbox_config_gen.py
swagner-de: I was actually thinking of something like that too, but figured since it is just 2 fields I would not benefit from it.

networking_ccloud/tools/netbox_config_gen.py
sebageek: Hmm. I see. Should we limit this to the two options then? I'm just thinking about somebody throwing an `blabla_address_scopes` in there and it breaks our config gen.

doc/source/configuration/config-gen.rst
```diff
@@ -17,12 +17,15 @@ It does support nesting the configuration under certain YAML keys if the config
 
 Vault references can also be injected instead of plain-text passwords using the :code:`--vault-ref` parameter.
 
+In our environment, we also need a mapping from OpenStack address-scopes to externally routable VRFs. We can supply that mapping using the :code:`--address-scope-vrf-map` parameter.
+Multiple mappings can be supplied. The Config Generator will check the YAML file for any top level key that ends in :code:`address_scopes` and will then try to parse a VRF and an address scope name out of it. For this to work, the list items must have a :code:`name` and :code:`vrf` key.
```
sebageek: This is now not completely right, as we only look at certain keys ;)

sebageek (approved):

PR #112, opened 2022-10-26
L3 shall rise again

sebageek:


networking_ccloud/common/config/config_driver.py
```diff
@@ -505,6 +506,13 @@ def get_availability_zone(self, az_name):
                 return az
         return None
 
+    def get_vrf_name_for_address_scope(self, address_scope):
```
swagner-de, line 523: Maybe we could annotate with an lru_cache? Also, we should maybe add a validator on the address_scope_name uniqueness.

networking_ccloud/db/db_plugin.py
```diff
@@ -270,3 +296,110 @@ def remove_transit_from_network(self, context, network_id, az):
...
+
+        # group by subnet pool
+        result = {}
+        for snp_id, cidr, net_az_hint in query.all():
+            az_hint = None
+            try:
+                if net_az_hint:
+                    net_az_hint = json.loads(net_az_hint)
+                    if len(net_az_hint) >= 1:
+                        az_hint = net_az_hint[0]
```
swagner-de, line 356: Do we make sure there will never be more than 1 az_hints? If so, why `>=` and not `==`?

networking_ccloud/db/db_plugin.py
```diff
@@ -270,3 +296,110 @@ def remove_transit_from_network(self, context, network_id, az):
...
+        # get az from tags
+        fields = [models_v2.SubnetPool.id, tag_models.Tag.tag]
+        query = context.session.query(*fields)
+        query = query.join(tag_models.Tag,
+                           models_v2.SubnetPool.standard_attr_id == tag_models.Tag.standard_attr_id)
+        query = query.filter(models_v2.SubnetPool.id.in_(subnetpool_ids))
+
+        snp_az = {}
+        for snp_id, tag in query.all():
+            if tag.startswith(cc_const.AZ_TAG_PREFIX):
```
swagner-de:
```
query = query.filter(tag_models.Tag.tag.like(f'{cc_const.AZ_TAG_PREFIX}%'))

```

networking_ccloud/ml2/agent/common/messages.py
```diff
@@ -169,13 +219,17 @@ class VlanTranslation(pydantic.BaseModel):
...
+    description: str = None
 
     native_vlan: pydantic.conint(gt=0, lt=4094) = None
     trunk_vlans: List[pydantic.conint(gt=0, lt=4094)] = None
     vlan_translations: List[VlanTranslation] = None
     portchannel_id: pydantic.conint(gt=0) = None
     members: List[str] = None
 
+    # vlan interfaces only
+    vrf: str = None
```
swagner-de: Shouldn't this be on the `VlanIface`?

networking_ccloud/ml2/agent/common/messages.py
```diff
@@ -355,7 +418,7 @@ def add_binding_host_to_config(self, hg_config, network_id, seg_vni, seg_vlan, t
                         else:
                             iface.native_vlan = seg_vlan
 
-    def add_segments(self, net_segments, top_segments):
+    def add_segments(self, net_segments, top_segments, net_gateways):
```
swagner-de: I think this shoud take a `vrf` paramter as well and push it through to the `add_binding_host_to_config`.

swagner-de (commented):
I think I am halfway through. I will continue Wednesday.

networking_ccloud/db/db_plugin.py
sebageek: Yeah, we have a `network_create_precommit()` hook that should prevent this. The `>=` is just there as a mean of "defensive programming", as I expect this to fail at some point because everything fails at some point and I'd not want the network to get regional then.

networking_ccloud/ml2/agent/common/messages.py
sebageek: Hmyeah, we don't use that anymore.

networking_ccloud/ml2/agent/common/messages.py
sebageek: vrf is now part of `net_gateways`

networking_ccloud/common/config/config_driver.py
```diff
@@ -233,6 +233,12 @@ def __hash__(self) -> int:
         return hash((self.name, self.vlan, self.vrf, tuple(self.networks),
                     self.vni, self.untagged, tuple(self.dhcp_relays)))
 
+    @pydantic.validator('networks')
+    def ensure_at_lest_one_address(cls, nets):
```
swagner-de: Objection, we defined Infra Networks as Layer 2 or Layer 3 objects. So an Infra Network without network and VRF bound could still be bound as flat L2 network.

networking_ccloud/common/config/config_driver.py
```diff
@@ -248,10 +254,6 @@ def ensure_network(cls, net):
 
     @pydantic.root_validator
     def ensure_correct_value_combination(cls, values):
```
swagner-de: On acceptance of the Layer 2 Infra Network comment, these validations should come back.

networking_ccloud/ml2/agent/common/messages.py
```diff
@@ -341,6 +399,11 @@ def add_binding_host_to_config(self, hg_config, network_id, seg_vni, seg_vlan, t
                 scu.add_vlan(seg_vlan, network_id)
                 scu.add_vxlan_map(seg_vni, seg_vlan)
 
+            # gateways
+            if gateways:
+                scu.add_vlan_iface(vlan=seg_vlan, vrf=gateways['vrf'], primary_ip=gateways['ips'][0],
```
swagner-de, line 413: We need at least a semi-stable way for the ordering of primary and secondary IPs. In this implmentation the order comes down all the way from the DB. With all the joins inbetween I don't think the order is stable. An unstable order leads to us swapping primary and secondary IP over and over, each time creating a hickup. We probably cannot prevent this completely but relying on the DB feels dangerous.

networking_ccloud/common/config/config_driver.py
sebageek: That sadly doesn't work with pydantic, so I opted for pregenerating the dicts. The problem was that pydantic no longer recognized the methods as methods.

networking_ccloud/common/config/config_driver.py
sebageek: You are right, I will bring that back.

networking_ccloud/ml2/agent/common/messages.py
sebageek: The method that fetches the gateways already sorts them by int value of their network ip, so we always have a stable ordering. Whenever the ordering changes we will reconfigure all ips and the ip with the lowest int value will be the primary ip.

PR #114, opened 2022-12-21
Agent prometheus monitoring

sebageek:
Currently on top of the l3 branch, will only be merged after l3 has been merged. Also, depends on https://github.com/sapcc/helm-charts/pull/4458 when deployed in our infra.


joker-at-work (approved):
I haven't implemented a prometheus exporter in Python, yet, but reading what you wrote here makes sense to me.

PR #119, opened 2023-02-24
WIP: Fix yoga tests

sebageek:
We need to configure MTUs, as else the networks are not getting created. The default MTU is 1500, for the VXLAN segment OpenStack substracts space for the header and our standard networks cannot be created anymore. Therefore I'm hardcoding the MTU in the tests.

Network segments need to be specified for the current allocation logic. Sadly we have a clash between two SegmentTestCases, where they both reconfigure the same DB. I've just put the same ranges into both cases, but a real fix would need separate DBs for both tests. This is why this commit is still WIP.

---

DO NOT MERGE, I'll to a proper one after my vacation.


sebageek:
Closed in favor of #121

PR #120, opened 2023-05-17
[netbox-model] Only log and ignore if name is inconsistent

swagner-de:
Whenever we encounter a device that does not fit out naming convention,
we raise an exception and terminate. That's tedious as we need to poke
people to correct all their switch names when we run this script.
Instead, we should just chose to ignore every device that does not
comply with the naming scheme and move on.


sebageek:
What is logged and ignored? Which names are inconsistent? Why do we need it / what problem does it solve?

sebageek (approved):

PR #121, opened 2023-05-26
Fix yoga tests

sebageek:
We need to configure MTUs, as else the networks are not getting created. The default MTU is 1500, for the VXLAN segment OpenStack substracts space for the header and our standard networks cannot be created anymore. Therefore I'm hardcoding the MTU in the tests.

Network segments need to be specified for the current allocation logic.

oslo.config's set_override() seems to not be cleaned up after a test is run. This means override of one test can influence tests run after this test. To avoid this we needed to add some cleanup statements for tenant_network_types. Before this commit the problematic behavior could be triggered with this command:

tox -e py38 -- --serial TestConfigValidation.test_validate_ml2_vlan_ranges_success networking_ccloud.tests.unit.db.test_db_plugin.TestDBPluginNetworkSyncData.test_get_hosts_on_segments_with_segment_with_wrong_level

### DIFF
--- a/networking_ccloud/tests/unit/common/test_config_validation.py
+++ b/networking_ccloud/tests/unit/common/test_config_validation.py
@@ -29,13 +29,11 @@ class TestConfigValidation(base.TestCase):
         super().setUp()
 
     def test_validate_ml2_vlan_ranges_success(self):
-        cfg.CONF.set_override('tenant_network_types', ['vxlan', 'vlan'], group='ml2')
         cfg.CONF.set_override('network_vlan_ranges', ['seagull:23:42', 'cat:53:1337'], group='ml2_type_vlan')
         cfg.CONF.set_override('driver_config_path', 'invalid/path/to/conf.yaml', group='ml2_cc_fabric')
         validate_ml2_vlan_ranges(self.conf_drv)
 
     def test_validate_ml2_vlan_ranges_failure(self):
-        cfg.CONF.set_override('tenant_network_types', ['vxlan', 'vlan'], group='ml2')
         cfg.CONF.set_override('network_vlan_ranges', ['cat:53:1337'], group='ml2_type_vlan')
         cfg.CONF.set_override('driver_config_path', 'invalid/path/to/conf.yaml', group='ml2_cc_fabric')
         self.assertRaisesRegex(cc_exc.MissingPhysnetsInNeutronConfig, ".*seagull.*",
--- a/networking_ccloud/tests/unit/db/test_db_plugin.py
+++ b/networking_ccloud/tests/unit/db/test_db_plugin.py
@@ -21,6 +21,7 @@ from neutron.services.tag import tag_plugin
 from neutron.services.trunk import models as trunk_models
 from neutron.tests.unit.extensions import test_segment
 from neutron_lib import context
+from neutron_lib.plugins import directory
 from oslo_config import cfg
 
 from networking_ccloud.common.config import _override_driver_config, config_driver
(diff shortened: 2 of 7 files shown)


networking_ccloud/tests/unit/common/test_config_validation.py
```diff
@@ -33,10 +33,12 @@ def test_validate_ml2_vlan_ranges_success(self):
         cfg.CONF.set_override('network_vlan_ranges', ['seagull:23:42', 'cat:53:1337'], group='ml2_type_vlan')
         cfg.CONF.set_override('driver_config_path', 'invalid/path/to/conf.yaml', group='ml2_cc_fabric')
         validate_ml2_vlan_ranges(self.conf_drv)
+        cfg.CONF.clear_override('tenant_network_types', group='ml2')
```
joker-at-work: Didn't you want to remove the `set_override()` instead? If not, this needs to be done differently, because the `clear_override()` needs to happen if any of the previous steps fail, too. It's currently written for a working test only.

networking_ccloud/tests/unit/db/test_db_plugin.py
```diff
@@ -34,6 +35,17 @@ class TestDBPluginNetworkSyncData(test_segment.SegmentTestCase, base.PortBinding
     def setUp(self):
         super().setUp()
 
+        cfg.CONF.set_override('global_physnet_mtu', 9000)
+        cfg.CONF.set_override('path_mtu', 9000, group='ml2')
+        cfg.CONF.set_override('network_vlan_ranges', ['foo:100:410', 'bar:200:210', 'baz:300:310',
+                                                      'spam:500:510', 'ham:600:610', 'mew:700:710',
+                                                      'caw:800:810'],
+                              group='ml2_type_vlan')
```
joker-at-work, line 43: If you introduce a config override, I think you should also have a cleanup for that.

networking_ccloud/tests/unit/extensions/test_extensions.py
```diff
@@ -117,6 +117,16 @@ def setUp(self):
         self.db = CCDbPlugin()
         self.ctx = context.get_admin_context()
 
+        cfg.CONF.set_override('global_physnet_mtu', 9000)
+        cfg.CONF.set_override('path_mtu', 9000, group='ml2')
+        cfg.CONF.set_override('network_vlan_ranges', ['seagull:100:1010', 'crow:200:210', 'bgw1:234:244',
+                                                      'bgw2:345:355', 'transit1:111:121', 'transit2:222:233'],
+                              group='ml2_type_vlan')
```
joker-at-work, line 124: I think this config override should be cleaned up.

networking_ccloud/tests/unit/extensions/test_extensions.py
```diff
@@ -215,13 +225,17 @@ def test_network_ensure_interconnects(self):
             self._make_segment(network_id=network_id, network_type='vxlan',
                                segmentation_id=424242,
                                tenant_id="test-tenant", fmt='json')['segment']
+            objs = self.ctx.session.query(segment_models.NetworkSegment).filter_by(physical_network=None,
+                                                                                   network_type='vxlan')
+            objs.update({'segment_index': 0})
```
joker-at-work, line 224: I don't get this part. Is it because you cannot specify the `segment_index` in `self._make_segment()`?

tox.ini
```diff
@@ -13,10 +13,12 @@ setenv =
    OS_STDOUT_CAPTURE=1
    OS_STDERR_CAPTURE=1
    OS_TEST_TIMEOUT=60
-deps = -c{env:UPPER_CONSTRAINTS_FILE:https://releases.openstack.org/constraints/upper/ussuri}
+deps = -c{env:UPPER_CONSTRAINTS_FILE:https://releases.openstack.org/constraints/upper/yoga}
```
joker-at-work, line 16: Wondering why this uses the upstream contraints instead of the sapcc ones.

sebageek: Good question. I think because that was part of the cookiecutter template I used when creating this.

networking_ccloud/tests/unit/db/test_db_plugin.py
sebageek: You mean for each time I have a config override somewhere? I think in general each module should override the things it needs to be a certain way and with that not much cleanup should be needed.

On the other hand I might still be a bit emotional about this, as I really expected config overrides to not be shared between tests.

networking_ccloud/tests/unit/extensions/test_extensions.py
sebageek:
```
{"NeutronError": {"type": "HTTPBadRequest", "message": "Unrecognized attribute(s) 'segment_index'", "detail": ""}}
```
`_make_segment()` doesn't accept this as an argument. If there is a way to do that I haven't looked hard enough yet.

joker-at-work: A comment to explain this, would have been nice, but not required.

joker-at-work (dismissed):

joker-at-work (approved):

joker-at-work (approved):

PR #122, opened 2023-06-07
Move GNMI client out of EOS agent, make it common

sebageek:
EOS was the only agent where we planned on using GNMI, but now it looks like we're going to use GNMI with NXOS as well. Therefore the GNMI client abstraction we have moves into common/ and all agents that want to use it can use it. The agents will keep their vendor specific stuff (like paths or regexes) in their own code base though.

### DIFF
--- a/networking_ccloud/ml2/agent/common/gnmi.py
+++ b/networking_ccloud/ml2/agent/common/gnmi.py
@@ -1,0 +1,132 @@
+# Copyright 2023 SAP SE
+#
+#    Licensed under the Apache License, Version 2.0 (the "License"); you may
+#    not use this file except in compliance with the License. You may obtain
+#    a copy of the License at
+#
+#         http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS, WITHOUT
+#    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
+#    License for the specific language governing permissions and limitations
+#    under the License.
+import json
+import logging
+import time
+
+from prometheus_client import Counter, Histogram
+from pygnmi.client import gNMIclient, gNMIException
+
+from networking_ccloud.common import exceptions as cc_exc
+
+
+LOG = logging.getLogger(__name__)
+
+
+class CCGNMIClient:
+    metrics_namespace = 'networking_ccloud_switch_agent_switch_gnmi'
+
+    DEFAULT_LABELS = ['platform', 'switch_name', 'switch_host', 'method']
+
+    metric_api_action = Histogram(
+        'api_action', 'GNMI API action duration',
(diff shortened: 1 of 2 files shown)


networking_ccloud/ml2/agent/common/gnmi.py
```diff
@@ -0,0 +1,132 @@
...
+            cmd = "".join(cmd)
+            LOG.debug("Failed command on %s was %s", self._switch_name, cmd)
+
+            if isinstance(e.orig_exc, AttributeError):
+                # module 'grpc' has no attribute '_channel', sometimes a first-connect problem
+                LOG.info("Reconnecting %s because of %s %s", self._switch_name, e.__class__.__name__, str(e))
+                self.connect()
+            if retries > 0:
+                self.metric_api_retry.labels(method=method, **self._def_labels).inc()
+                time.sleep(0.5)
```
swagner-de, line 103: How about an exponential backoff like `time.sleep(2 ** retry)`?

networking_ccloud/ml2/agent/common/gnmi.py
```diff
@@ -0,0 +1,132 @@
...
+
+            def _unpack(entry):
+                if single:
+                    return entry[0]['val']
+                else:
+                    return [x['val'] for x in entry]
+
+            data = [_unpack(e) for e in data]
+
+            if not (path and len(path) > 1):
```
swagner-de, line 126: I feel like, I would expect to always become a list of lists back when `single=False`, independent if the inner list has just one item. This way `single` is only respected when `unpack=True`. I know you pin it on the path length, yet I just assume that in some cases you generate the paths dynamically (maybe like a list of interfaces or bgp networks) and if those happen to be just one, you get a different data type back.

swagner-de (approved):
Since this has already been productive code and this is only moved into a different class, approving with comments.

networking_ccloud/ml2/agent/common/gnmi.py
sebageek: If you want to have this feature I'd be open to a PR, but wouldn't implement this here where I just want to refactor code.

networking_ccloud/ml2/agent/common/gnmi.py
sebageek: We have four levels to unpack:
 * Level zero is the outside `notification` business, controlled by `unpack=True`
 * Level one is "one answer per path". If multiple paths are requested we return a list with one entry for each requested path. For a single path we do automatic unpacking. If you want we can add an extra switch for this to turn off this behavior. We can also introduce it later, once it is needed.
 * Level two / `single=True` refers to how many answers a single path is getting. pygnmi always returns a list but in many cases we only have one element that we're interested is, so this is again to make it easier to work with the API responses
 * Level Three is the inside extra `val` dictionary key

So if we need an extra control for "level one" we could add it, but I'd also think we should do this in an extra PR.

PR #124, opened 2023-07-19
Nxos agent base

sebageek:
This is the basic agent for configuring OpenStack tenant networks on NXOS switches. The PR still contains a lot of FIXMEs and rough edges. I decided to leave them in so we would have this preliminary version running in our QA landscape without having to depend on me doing temporary deployments.

### DIFF
--- a/networking_ccloud/ml2/agent/nxos/switch.py
+++ b/networking_ccloud/ml2/agent/nxos/switch.py
@@ -11,88 +11,483 @@
 #    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
 #    License for the specific language governing permissions and limitations
 #    under the License.
-
-import time
+import re
 
 from oslo_log import log as logging
-import requests
+from typing import List, Optional
 
 from networking_ccloud.common import constants as cc_const
+from networking_ccloud.ml2.agent.common.gnmi import CCGNMIClient
+from networking_ccloud.ml2.agent.common import messages as agent_msg
+from networking_ccloud.ml2.agent.common.messages import OperationEnum as Op
 from networking_ccloud.ml2.agent.common.switch import SwitchBase
 
 LOG = logging.getLogger(__name__)
 
+# NXOS GNMI limitations
+# https://www.cisco.com/c/en/us/td/docs/switches/datacenter/nexus9000/sw/93x/progammability/guide/b-cisco-nexus-9000-series-nx-os-programmability-guide-93x/b-cisco-nexus-9000-series-nx-os-programmability-guide-93x_chapter_0110001.html
+# https://github.com/openconfig/reference/blob/1cf43d2146f9ba70abb7f04f6b0f6eaa504cef05/rpc/gnmi/gnmi-specification.md
+
+
+class NXOSGNMIPaths:
+    IFACE_VTRANS_ITEM = ("System/intf-items/phys-items/PhysIf-list[id={iface}]/vlanmapping-items/"
+                         "vlantranslatetable-items/vlan-items/VlanTranslateEntry-list[vlanid=vlan-{inside}]"
+                         "[translatevlanid=vlan-{outside}]")
+    IFACE_PC_VTRANS_ITEM = ("System/intf-items/aggr-items/AggrIf-list[id={iface}]/vlanmapping-items/"
+                            "vlantranslatetable-items/vlan-items/VlanTranslateEntry-list[vlanid=vlan-{inside}]"
+                            "[translatevlanid=vlan-{outside}]")
+
+
+class NXOSSetConfig:
+    def __init__(self):
+        self.delete = []
+        self.replace = []
+        self.update = []
+
+    def get_list(self, op):
+        if op == Op.add:
+            return self.update
+        elif op == Op.replace:
+            return self.replace
+        else:
+            raise ValueError("Only available for add/replace")
+
 
 class NXOSSwitch(SwitchBase):
+    # path is System/intf-items/phys-items/PhysIf-list[id=eth1/17]/trunkVlans
+    #         System/intf-items/aggr-items/AggrIf-list[id=po6]/trunkVlans
+    IFDN_RE = re.compile(r"^System/intf-items/(?:aggr-items/AggrIf|phys-items/PhysIf)-list\[id=(?P<ifname>[^\]]+)]")
+
     @classmethod
     def get_platform(self):
         return cc_const.PLATFORM_NXOS
 
     def login(self):
-        self._api = requests.Session()
-        self._api.auth = (self.user, self._password)
-
-    @classmethod
-    def _to_payload(cls, cmds, format_text):
-        if isinstance(cmds, str):
-            cmds = [cmds]
-
-        payload = []
-        for n, cmd in enumerate(cmds):
-            pl = {
-                "jsonrpc": "2.0",
-                "method": "cli" if not format_text else "cli_ascii",
-                "params": {
-                    "cmd": cmd,
-                    "version": 1,
-                },
-                "id": n,
-            }
-            payload.append(pl)
-        return payload
-
-    def send_cmd(self, cmd, format_text=False, raw=False, _is_retry=False):
-        """Send a command to the switch"""
-
-        headers = {
-            'content-type': 'application/json-rpc'
-        }
-        start_time = time.time()
-        try:
-            payload = self._to_payload(cmd, format_text=format_text)
-            resp = self.api.post(f"https://{self.host}/ins", headers=headers, json=payload, verify=self._verify_ssl)
-            # FIXME: do we want to raise for status?
-        except Exception:
-            LOG.exception("Command failed in %.2fs on %s %s, cmd: %s",
-                          time.time() - start_time, self.name, self.host, cmd)
-            raise
-
-        LOG.debug("Command succeeded in %.2fs on %s %s, cmd: %s", time.time() - start_time, self.name, self.host, cmd)
-        result = resp.json()
-
-        if not raw:
-            # unpack the response a bit for easier handling
-            if not isinstance(result, list):
-                result = [result]
-
-            # FIXME: will there always be a result? No, not if the command failed. should we raise then?
-            for n, entry in enumerate(result):
-                if 'result' in entry:
-                    result[n] = entry['result']['body']
-
-            # unpack the response if user only specified a string as cmd
-            if isinstance(cmd, str):
-                result = result[0]
-
-        return result
+        self._api = CCGNMIClient(switch_name=self.name, host=self.host, port=50051,
+                                 username=self.user, password=self._password, platform=self.get_platform(),
+                                 insecure=False, skip_verify=True)
+        self._api.connect()
 
     def _get_switch_status(self):
-        ver = self.send_cmd("show version")
+        fw_ver, model, uptime = self.api.get(path=[
+            "/System/showversion-items/nxosVersion",
+            "/System/ch-items/model",
+            "/System/showversion-items/kernelUptime"])
         return {
             'name': self.name,
             'host': self.host,
             'api_user': self.user,
-            'version': ver['nxos_ver_str'],
-            'model': ver['chassis_id'],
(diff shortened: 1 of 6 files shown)


networking_ccloud/ml2/agent/nxos/switch.py
```diff
@@ -11,88 +11,839 @@
...
-            if isinstance(cmd, str):
-                result = result[0]
 
-        return result
+class NXOSSwitch(SwitchBase):
+    # path is System/intf-items/phys-items/PhysIf-list[id=eth1/17]/trunkVlans
+    #         System/intf-items/aggr-items/AggrIf-list[id=po6]/trunkVlans
+    IFDN_RE = re.compile(r"^System/intf-items/(?:aggr-items/AggrIf|phys-items/PhysIf)-list\[id=(?P<ifname>[^\]]+)]")
+    # /System/inst-items/Inst-list[name='CC-CLOUD01']
+    VRF_TDN_RE = re.compile(r"^/System/inst-items/Inst-list\[name='(?P<vrf>[^']+)'\]$")
```
sven-rosenzweig, line 71: Is the ``'`` in the path necessary? 
looking at the ``VRF_TDN_RE`` and ``BGP_VRF_RE`` paths seem not consistent to me. 
``Inst-list[name='CC-CLOUD01']`` vs. ``[name=CC-CLOUD01].``

networking_ccloud/ml2/agent/nxos/switch.py
```diff
@@ -11,88 +11,839 @@
...
 
-class NXOSSwitch(SwitchBase):
-    @classmethod
-    def get_platform(self):
-        return cc_const.PLATFORM_NXOS
 
-    def login(self):
-        self._api = requests.Session()
-        self._api.auth = (self.user, self._password)
+def guess_asn_format(asn):
```
sven-rosenzweig, line 35: Do we have a test for the method? Do we want one?

networking_ccloud/ml2/agent/nxos/switch.py
```diff
@@ -11,88 +11,839 @@
...
+    OPT_GNMI_RUN_SEPARATE = False
+    GNMI_CHUNK_SIZE = 20  # NXOS supports a maximum of 20 paths per request
+    TRUNCATE_VLAN_NAMES = False
+
+    @classmethod
+    def get_platform(cls):
+        return cc_const.PLATFORM_NXOS
+
+    def login(self):
+        self._api = CCGNMIClient(switch_name=self.name, host=self.host, port=50051,
```
sven-rosenzweig: Any specific reason why the port is hardcoded? Do we benefit from making this configurable?

networking_ccloud/ml2/agent/nxos/switch.py
```diff
@@ -11,88 +11,839 @@
...
+                                 username=self.user, password=self._password, platform=self.get_platform(),
+                                 insecure=False, skip_verify=True)
+        self._api.connect()
```
sven-rosenzweig, line 94: What happens if a login fails? Where do we handle the exception?

networking_ccloud/ml2/agent/nxos/switch.py
```diff
@@ -11,88 +11,839 @@
...
+        # (the base driver code should always take care of this)
+        if not (vlans and vxlan_maps):
+            return
+
+        vlan_vxmaps = {vx.vlan: vx.vni for vx in vxlan_maps}
+        if {v.vlan for v in vlans} ^ set(vlan_vxmaps):
+            LOG.warning("Inconsistent config request: Vlan IDs %s configured alongside vlan <-> vxlan maps %s on %s",
+                        [v.vlan for v in vlans], list(vlan_vxmaps), self)
+
+        if operation in (Op.add, Op.replace):
```
sven-rosenzweig, line 159: Quite a big if/else. This would be my counter suggestion: 
```
if operation == Op.remove
    for v in vlans:
            config_req.delete.append(f"/System/bd-items/bd-items/BD-list[fabEncap=vlan-{v.vlan}]")
            for vx in vxlan_maps:
                config_req.delete.append("/System/eps-items/epId-items/Ep-list[epId=1]/"
                                         f"nws-items/vni-items/Nw-list[vni={vx.vni}]")
             return


if operation == Op.replace:
    # remove vlans managed by us that are not part of our mappings
     # (this is done as an extra block as we want to delete the vlan even if there's no vni mapped to it)
     vlans_to_remove = self.get_all_managed_vlan_ids_on_switch() - {v.vlan for v in vlans}
     for vlan in sorted(vlans_to_remove):
                    LOG.debug("Removing stale vlan %s from %s on config replace", vlan, self)
                    config_req.delete.append(f"/System/bd-items/bd-items/BD-list[fabEncap=vlan-{vlan}]")
     .....

# Add or replace operation
 # vlan part
  all_vlans = []
  for v in vlans:
       # create the vlan
       vlan_name = v.name
    

``

File: networking_ccloud/ml2/agent/nxos/switch.py (hunk 3 above)
Comment (felix-kaestner) [reply networking_ccloud/ml2/agent/nxos/switch.py]: We also use port `9339` for the gnmi api by those devices bootstrapped with the network-operator. So I guess making it configurable would help.

File: networking_ccloud/ml2/agent/nxos/switch.py (hunk 1 above)
Comment (felix-kaestner) [inline networking_ccloud/ml2/agent/nxos/switch.py:71]: While it doesn't seem to make a difference, we sometimes prefix the YANG path with a `/` as in `/System` and sometimes we don't. Should we align this to have one way or the other consistently?

File: networking_ccloud/ml2/agent/nxos/switch.py [hunk 6]
```diff
@@ -11,88 +11,839 @@
...
-            'version': ver['nxos_ver_str'],
-            'model': ver['chassis_id'],
-            'uptime': ver['rr_ctime'],  # FIXME: convert to seconds since start
+            'version': fw_ver,
+            'model': model,
+            'uptime': uptime,  # FIXME: convert to seconds since start
         }
+
+    def get_vlan_and_vxmap_config(self) -> Tuple[List[agent_msg.Vlan], List[agent_msg.VXLANMapping]]:
+        swdata = self.api.get(path=["/System/bd-items/bd-items"])['BD-list']
```
felix-kaestner, line 111: What was the reason to choose this over:
```suggestion
        swdata = self.api.get(path=["/System/bd-items/bd-items/BD-list"], single=False)
```

networking_ccloud/ml2/agent/nxos/switch.py
```diff
@@ -11,88 +11,839 @@
...
+        for data in (ifdata + pcdata):
+            if data.get('rtmbrIfs-items'):
+                # this interface is part of a portchannel, skipping reporting its config
+                continue
+
+            iface = agent_msg.IfaceConfig(name=data['id'])
+            if data.get('descr'):
+                iface.description = data['descr']
+            if data.get('nativeVlan'):
+                iface.native_vlan = data['nativeVlan'][len("vlan-"):]
```
felix-kaestner:
```suggestion
                iface.native_vlan = int(data['nativeVlan'][len("vlan-"):])
```
should this be casted to an int?

networking_ccloud/ml2/agent/nxos/switch.py
```diff
@@ -11,88 +11,839 @@
...
+            for vx in vxlan_maps:
+                nve_item = {
+                    'vni': vx.vni,
+                    'suppressARP': 'enabled' if vx.vlan in svi_vlans else 'off',
+                    # 'IngRepl-items': {'proto': 'bgp'},
+                    'IngRepl-items': '',
+                    'multisiteIngRepl': 'enable' if vx.enable_multisite else 'disable',
+                }
+                nve_list.append(nve_item)
+            config_req.update.append(("/System/eps-items/epId-items/Ep-list[epId=1]/nws-items/vni-items",
```
felix-kaestner, line 218: I think we need to verify that NX-OS really treats this as a merge operation on that path, only adding/patching items that are provided in the request and NOT deleting existing ones not provided in the request. Otherwise we might accidentally delete internal VNIs managed by the network-operator.

networking_ccloud/tests/unit/ml2/agent/nxos/test_switch.py
```diff
@@ -0,0 +1,378 @@
...
+    def setUp(self):
+        super().setUp()
+        drv_conf = cfix.make_config(global_config=cfix.make_global_config(asn_region=65130))
+        _override_driver_config(drv_conf)
+        cfg_switch = config_driver.Switch(name="seagull-sw1", host="127.0.0.1", platform=cc_const.PLATFORM_NXOS,
+                                          user="seagulladm", password="KRAKRAKRA", bgp_source_ip="1.1.1.1")
+
+        cfg.CONF.set_override('prometheus_enabled', False, group='ml2_cc_fabric_agent')
+        with mock.patch.object(n_rpc, 'get_client'):
+            self.switch = NXOSSwitch(cfg_switch, 65130, 'a', {100} | set(range(2000, 3000)), [range(10000, 11000)],
```
sven-rosenzweig: I think ``NXOSSwitch(sw_conf=cfg_switch, asn_region='a', ...``would increase understandability.

networking_ccloud/ml2/agent/nxos/switch.py
sebageek: I took this style from what I receive when I query the API:
```
>>> sw.grpc_get(path=["/System/intf-items/svi-items/If-list/rtvrfMbr-items"])
<<< {'tDn': "/System/inst-items/Inst-list[name='CC-CLOUD01']"}
```
I don't have specific reason to do so expect reproducing what the API returns. Should we investigate here?

networking_ccloud/ml2/agent/nxos/switch.py
sebageek: If connection fails we raise an exception and that will be logged. We do retries inside the CCGNMIClient, but we don't save the request when all retries are exceeded. It'll then need to be handled by the next full sync.

networking_ccloud/ml2/agent/nxos/switch.py
sebageek: I tested both and it looks like tune single=False combined with unpack at the moment seems to wrap an extra layer of lists around list elements. I haven't explored this further, but it seems to be making a difference.

networking_ccloud/ml2/agent/nxos/switch.py
sebageek:
```
(qa-de-1) [58] >>> sw.grpc.set(update=[("/System/eps-items/epId-items/Ep-list[epId=1]/nws-items/vni-items", {"Nw-list": [{"vni": 133737}]})])
(qa-de-1) [58] <<< 
{'timestamp': 1786111002871183916,
 'prefix': None,
 'response': [{'path': 'System/eps-items/epId-items/Ep-list[epId=1]/nws-items/vni-items',
   'op': 'UPDATE'}]}
```
this merges the list of vnis. all previously present vnis stay there, only the new `133737` vni is added.

networking_ccloud/ml2/agent/nxos/switch.py
sebageek: Hmmm, I have the same structure in every method so far: if op in (add, replace): handle them (cause they are generally somewhat similar) else do the delete part. I'd go with consistency here, what do you think?

networking_ccloud/ml2/agent/nxos/switch.py
sven-rosenzweig: No, I do not see the benefit of this investigation. this just popped into my eyes and I added this as feedback to this PR.

networking_ccloud/ml2/agent/nxos/switch.py
sven-rosenzweig: the existing pattern is fine by me.

networking_ccloud/ml2/agent/nxos/switch.py
felix-kaestner: I see. This actually seems to be a mistake in how Cisco NX-OS marshals the payload. As it seems to be that a single tree (like the "BD-list") is wrapped in an additional array for now reason.

PR #128, opened 2023-10-06
Netbox 3.0 adjustments

swagner-de:
With netbox 3.0, there were 2 API changes that affect us:
1. Interfaces can now carry multiple connections, I believe is that is
   to support broken out interfaces. We do not plan to have break outs
   in the future, however we have them now. Yet we model them in the
   form that we create a new interface (which could carry individual
   VLANs and so on) and then model the connection. Hence, there is no
   need for us to support multiple connected endpoints under an
   interface. If we encounter it, we fail.

2. Sites can have multiple ASNs now, ASNs are objects, not plain ints
   anymore. This is not an issue so far as a site currently only carries
   one ASN, which we simply unpack. Yet, I can imagine that we use
   multiple ASN in a site in the future in which case we would currently
   throw an exceptions. That's fine for now. If the time comes, we
   probably need to tag the ASN we want to use for the fabric.

### DIFF
--- a/networking_ccloud/tools/netbox_config_gen.py
+++ b/networking_ccloud/tools/netbox_config_gen.py
@@ -486,7 +486,7 @@ class ConfigGenerator:
 
     def get_asn_region(self, region):
         sites = self.netbox.dcim.sites.filter(region=region)
-        site_asns = {site.asn for site in sites if site.asn}
+        site_asns = {asn for site in sites if site.asns for asn in site.asns}
         if not site_asns:
             raise ConfigException(f"Region {region} has no ASN")
         if len(site_asns) > 1:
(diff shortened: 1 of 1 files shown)


networking_ccloud/tools/netbox_config_gen.py
```diff
@@ -486,7 +477,10 @@ def get_l3_data(self, asn_region: int, pod: int, switchgroup_no: int, leaf_no: i
 
     def get_asn_region(self, region):
         sites = self.netbox.dcim.sites.filter(region=region)
-        site_asns = {site.asn for site in sites if site.asn}
+        site_asns = set()
+        for site in sites:
+            if site.asns:
+                site_asns.update(x.asn for x in site.asns)
```
sebageek: Optional: Ooooor we do a double-for :D 
```python
site_asns = {asn for site in sites if site.asns for asn in site.asns}
```
Tried it that it works this way:
```python
In [1]: a = [[1,2,3], None, [3,2,1]]                                                                                                                                                  

In [2]: [y for x in a if x for y in x]                                                                                                                                                
Out[2]: [1, 2, 3, 3, 2, 1]
```

swagner-de: Yep, that looks good.

sebageek (approved):

PR #131, opened 2023-10-30
Implement trunk extension

sebageek:
Implement the trunk service extension. A trunk can be created on any port that is a in a direct binding hostgroup. Only one trunk per hostgroup is possible. If the host is part of a metagroup, only VLANs not used by the metagroup VLAN pool can be used as target, to not interfere with VMs put on the host(s).

The trunk plugin modifies the subports by setting binding host, vnic type and binding profile. The binding profile contains details about the specific VLAN translation, but this is only informational, as the "real" info will be fetched from the respective trunk/subport tables in the DB.


networking_ccloud/services/trunk/driver.py
```diff
@@ -0,0 +1,227 @@
...
+                    'device_id': trunk_port.get('device_id'),
+                }
+            }
+            self.core_plugin.update_port(context, subport.port_id, port_data)
+        if len(subports) > 0:
+            trunk.update(status=trunk_const.TRUNK_ACTIVE_STATUS)
+
+    def _unbind_subports(self, context, trunk_port, trunk, subports):
+        for subport in subports:
+            LOG.info("Adding subport %s trunk port %s of trunk %s", subport.port_id, trunk.port_id, trunk.id)
```
joker-at-work: I think this needs to be "Removing". Looks like a copy&paste error.

networking_ccloud/services/trunk/driver.py
```diff
@@ -0,0 +1,227 @@
...
+                    'device_id': '',
+                    'status': nl_const.PORT_STATUS_DOWN,
+                },
+            }
+            self.core_plugin.update_port(context, subport.port_id, port_data)
+
+            if len(trunk.sub_ports) - len(subports) > 0:
+                trunk.update(status=trunk_const.TRUNK_ACTIVE_STATUS)
+            else:
+                LOG.info("Last subport was removed from trunk %s, setting it to state DOWN", trunk.id)
```
joker-at-work, line 228: You do not set to state down here. Does that happen automatically and this is just an FYI?

Should this rather be `trunk_const.TRUNK_DOWN_STATUS` or something therelike?

Does updating the trunk still work if we run this code in `AFTER_DELETE` for the trunk?

sebageek: After an operation of the trunk it is automatically set to down if we don't do anything against that.

sven-rosenzweig (approved):
Looks good to me. I like the test driven approach here.

PR #136, opened 2023-12-12
[config-gen] Add interface speed from netbox

swagner-de:
We need to force the speed on some breakout/QSA interfaces. As
port-channel members are just plain strings, we add the speed to the
port-channel interfaces itself. That is fine, as members can only be of
the same speed anyway. For physical interfaces, we can annotate it
straight to the interface.

### DIFF
--- a/networking_ccloud/tools/netbox_config_gen.py
+++ b/networking_ccloud/tools/netbox_config_gen.py
@@ -82,6 +82,9 @@ class ConfigGenerator:
     }
     ignore_tags = {"cc-net-driver-ignore"}
     extra_vlan_tag = "cc-net-driver-extra-vlan"
+    force_speed_tag = 'cc-net-driver-force-speed'
+    re_speed_tag = re.compile(r'^cnd-net-portspeed-(?P<speed>\d+g)$')
+    lag_member_speed_cache: Dict[int, str] = dict()
 
     # metagroup handlers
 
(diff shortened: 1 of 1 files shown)


networking_ccloud/tools/netbox_config_gen.py
```diff
@@ -82,6 +82,9 @@ class ConfigGenerator:
     }
     ignore_tags = {"cc-net-driver-ignore"}
     extra_vlan_tag = "cc-net-driver-extra-vlan"
+    force_speed_tag = 'cc-net-driver-force-speed'
+    speed_tag_regex = re.compile(r'^cnd-net-portspeed-(?P<speed>\d+g)$')
```
sebageek: All other regexes end with `_re` (see above). Do you want to do that here as well?

networking_ccloud/tools/netbox_config_gen.py
```diff
@@ -296,6 +329,7 @@ def handle_interface_or_portchannel(cls, iface: NbRecord, candidate_interfaces:
                                         switchport_kwargs: Dict[str, Any] = {}):
 
         device_name = iface.device.name  # type: ignore
+        speed = cls._handle_interface_speed(iface)
```
sebageek, line 331: even if it's a class method I'd use `self` here to call it.

networking_ccloud/tools/netbox_config_gen.py
```diff
@@ -213,6 +216,36 @@ def _ensure_single_item(cls, item_set: Iterable[Any], entity: str, identifier: s
                                         f'Unexpected values for {attribute}, expected 1 but got {len(item_set)}')
         return item_set.pop()
 
+    @classmethod
+    def _handle_interface_speed(cls, iface: NbRecord) -> Optional[str]:
+        force_speed = cls.force_speed_tag in {x.slug for x in iface.tags}
+        if not force_speed:
+            return None
```
sebageek:
```python
if not any(tag.slug == cls.force_speed_tag for tag in iface.tags):
    return None
```
This would only iterate over the list until the first matching item is found, though performance impact would be negligible with the short lists that we have.

networking_ccloud/tools/netbox_config_gen.py
```diff
@@ -213,6 +216,36 @@ def _ensure_single_item(cls, item_set: Iterable[Any], entity: str, identifier: s
...
+
+        lag = iface.lag
+        if lag.id not in cls.lag_member_speed_cache:
+            cls.lag_member_speed_cache[lag.id] = speed  # type: ignore
+            return speed
+
+        if cls.lag_member_speed_cache[lag.id] != speed:  # type: ignore
+            raise ConfigException(f'Interface {iface.id} has a different speed than other '
+                                  f'member interface(s) of LAG {lag.id}, has {speed}, other(s) have '
+                                  f'{cls.lag_member_speed_cache[lag.id]}')  # type: ignore
```
sebageek: Is `lag.id` enough to identify the port channel or would it be easier to the user to have a devicename (or switchport) + port name to find the offending entity?

sebageek (commented):
Code looks good so far, also tested it and it works. Added a couple of comments.

networking_ccloud/tools/netbox_config_gen.py
swagner-de: I am calling it from a classmethod. How'd I have `self` available?

networking_ccloud/tools/netbox_config_gen.py
swagner-de: Agreed.

networking_ccloud/tools/netbox_config_gen.py
swagner-de: make sense

networking_ccloud/tools/netbox_config_gen.py
swagner-de: I mean lag.id is the primary key in the database table and netbox won't create any duplicate `(device_name, interface_name)`. I wouldn't see a reason to use anything else.

networking_ccloud/tools/netbox_config_gen.py
sebageek: Whops, you wouldn't of course. Seems like I looked wrong.

networking_ccloud/tools/netbox_config_gen.py
sebageek: I was just thinking about the error message displayed to the user. Using `lag.id` as dict key totally makes sense to me.

sebageek (dismissed):
Error message with extra details is kinda optional for me

sebageek (approved):

PR #139, opened 2024-04-08
Support loading credentials from extra file

sebageek:
To allow the user of our driver to inject credentials from different sources we now support to load the credentials from an extra file. If the driver_config_credentials_path is defined, the driver will load it and patch credentials used inside the driver config where appropriate.

This was implemented to load credentials from a Kubernetes secret, while keeping the rest of the config inside a config map.


sebageek (commented):
Reviewed @swagner-de's changes and if it weren't my PR I'd approve of them!

swagner-de (approved):

sven-rosenzweig (approved):

PR #140, opened 2024-04-24
Untagged infra networks

sebageek:


sven-rosenzweig (approved):
fine by me. we could think about testing if ` add_infra_networks_from_hostgroup(self, hg_config, sg, process_untagged=False)` is called with correct parameters.

defo89 (approved):

mutax (approved):
lgtm, seagull and herring approved

PR #142, opened 2024-08-19
Nxos agent config preparations

sebageek:
This is everything we need to get the NXOS config rolling for the fabric driver (at least everything I know of) and also supports mixed-vendor switchgroups, so we can act as if multiple EVPN domains are one in our qa landscape. I also included the multisite part, so we'll only need to patch the agent in most cases when we're continuing development on the NXOS agent.


occamshatchet (approved):
Looks good and nice work!

mutax (approved):

PR #145, opened 2024-12-12
EOS GNMI compatibility for EOS >= 4.32

sebageek:
Somewhere between EOS 4.28.9M and 4.32.3M the GNMI structure for vlan-vni mappings and for aggregates changed.

For vlan-vni mappings the path changed, so we now need to check on which version we are and use the corresponding path for querying and configuring them. It's a bit weird now as each vlan element has a config and a state entry, but it looks like EOS just wants to introduce more state reporting into their GNMI API.

For the aggregates, we now need to filter both for protocol name and identifier name. In case of BGP they're both BGP. Querying works even without specifying the identifier, but configuration does not work anymore. Adding [identifier=BGP] to all of them solves this problem.

With these two changes we now need to implement a version switch. For now the version is fetched on first use and gets reset on each connect, so we make sure that we pick up new versions if we loose the GNMI connection (as this could very well be caused by a switch firmware update!). The version check is a rather simple tuple conversion / comparison and is currently based on the assumption that the first two parts of the version number will always be plain numeric (like 4.32.3M).

### DIFF
--- a/networking_ccloud/ml2/agent/eos/switch.py
+++ b/networking_ccloud/ml2/agent/eos/switch.py
@@ -40,8 +40,13 @@ class EOSGNMIPaths:
     VLAN = "network-instances/network-instance[name=default]/vlans/vlan[vlan-id={vlan}]"
 
     VXMAPS = "interfaces/interface[name=Vxlan1]/arista-exp-eos-vxlan:arista-vxlan/config/vlan-to-vnis"
+    VXMAPS_ALL_4_32 = ("interfaces/interface[name=Vxlan1]/arista-exp-eos-vxlan:arista-vxlan/vlan-to-vnis/"
+                       "vlan-to-vni/config")
+    VXMAPS_CONFIG_4_32 = "interfaces/interface[name=Vxlan1]/arista-exp-eos-vxlan:arista-vxlan/vlan-to-vnis"
     VXMAP_VLAN = ("interfaces/interface[name=Vxlan1]/arista-exp-eos-vxlan:arista-vxlan/config/vlan-to-vnis"
                   "/vlan-to-vni[vlan={vlan}]")
+    VXMAP_VLAN_4_32 = ("interfaces/interface[name=Vxlan1]/arista-exp-eos-vxlan:arista-vxlan/vlan-to-vnis"
+                       "/vlan-to-vni[vlan={vlan}]")
     VRF_VXMAPS = "interfaces/interface[name=Vxlan1]/arista-exp-eos-vxlan:arista-vxlan/config/vrf-to-vnis"
     VRF_VXMAP_VRF = ("interfaces/interface[name=Vxlan1]/arista-exp-eos-vxlan:arista-vxlan/config/vrf-to-vnis/"
                      "vrf-to-vni[vrf={vrf}]")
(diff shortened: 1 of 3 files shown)


toanju:
Looking at the yang models the change seem to have happend in 4.32:

https://github.com/aristanetworks/yang/blob/master/EOS-4.31.0F/experimental/eos/models/vxlan/arista-exp-eos-vxlan-config.yang

https://github.com/aristanetworks/yang/blob/master/EOS-4.32.0F/experimental/eos/models/vxlan/arista-exp-eos-vxlan-config.yang#L417

networking_ccloud/ml2/agent/eos/switch.py
```diff
@@ -105,11 +111,30 @@ class EOSSwitch(SwitchBase):
...
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+
+        self._reset_version_detection()
+
+    def _reset_version_detection(self):
+        self._version_min_4_32 = None
+
+    @property
+    def version_min_4_35(self):
```
ikatza-at-work: why is the function called `version_min_4_35`? this is not obvious...

networking_ccloud/ml2/agent/eos/switch.py
```diff
@@ -40,22 +40,28 @@ class EOSGNMIPaths:
     VLAN = "network-instances/network-instance[name=default]/vlans/vlan[vlan-id={vlan}]"
 
     VXMAPS = "interfaces/interface[name=Vxlan1]/arista-exp-eos-vxlan:arista-vxlan/config/vlan-to-vnis"
+    VXMAPS_ALL_4_32 = ("interfaces/interface[name=Vxlan1]/arista-exp-eos-vxlan:arista-vxlan/vlan-to-vnis/"
+                       "vlan-to-vni/config")
+    VXMAPS_CONFIG_4_32 = "interfaces/interface[name=Vxlan1]/arista-exp-eos-vxlan:arista-vxlan/vlan-to-vnis/"
     VXMAP_VLAN = ("interfaces/interface[name=Vxlan1]/arista-exp-eos-vxlan:arista-vxlan/config/vlan-to-vnis"
                   "/vlan-to-vni[vlan={vlan}]")
+    VXMAP_VLAN_4_32 = ("interfaces/interface[name=Vxlan1]/arista-exp-eos-vxlan:arista-vxlan/vlan-to-vnis"
+                       "/vlan-to-vni[vlan={vlan}]")
```
ikatza-at-work: Will we ever supporting more versions simultaneously?  This will obviously work, but may be doing encoding these paths in a dict keyed by the version might be an option if we ever go for that cases (probably less performant on the other hand).

ikatza-at-work: Looking at the code below that doesn't solve the problem many if-else blocks you would still have though

networking_ccloud/ml2/agent/eos/switch.py
sebageek: We're in the switch class and the idea was that with this property we have a flag to check if the switch version we can expect is at minimum 4.35. I roughly copied this from what I did in the asr1k driver, but I'd be open for other naming ideas. Do you have any at hand that you think could be more obvious?

networking_ccloud/ml2/agent/eos/switch.py
sebageek: Yeah, we most likely will need to support different versions, but most of the paths will not be keyed to a specific version. It will most likely be "before breaking the path" and "after breaking the path". In some cases we also need to revise the logic of how we construct the data, so it will most likely stay flag-based or we would need to write different methods that create certain parts of the config (maybe with decorators deciding for which version they are?).

networking_ccloud/ml2/agent/eos/switch.py
ikatza-at-work: would `version_gt_4_32` break anything? It's a bit odd to use the version in 4.35 in the header but then the implementation inside refers to 4_32

networking_ccloud/ml2/agent/eos/switch.py
ikatza-at-work: Yeah, that sounds reasonable. Probably not something strictly necessary for this PR

networking_ccloud/ml2/agent/eos/switch.py
sebageek: Ooooh, I completely overlooked the 4.35 vs 4.32 part. I just mistyped the version in the header and reused the name. I'll fix it.

toanju (approved):
lgtm

mutax (approved):
lgtm

sven-rosenzweig (approved):
LGTM

PR #150, opened 2025-07-31
Common agent backports nxos preparation

sebageek:
Starting to backport some of the patches needed for the nxos agent.

### DIFF
--- a/networking_ccloud/ml2/agent/common/agent.py
+++ b/networking_ccloud/ml2/agent/common/agent.py
@@ -18,6 +18,7 @@ import time
 
 from neutron.common import config as common_config
 from neutron.conf.agent.common import register_agent_state_opts_helper
+from neutron.conf.plugins.ml2.drivers.driver_type import register_ml2_drivers_vxlan_opts
 from neutron import manager
 from oslo_config import cfg
 from oslo_log import log as logging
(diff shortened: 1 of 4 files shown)


toanju (approved):

networking_ccloud/ml2/agent/common/agent.py
```diff
@@ -90,17 +91,32 @@ def __init__(self, *args, **kwargs):
...
 
+    def _get_managed_vnis(self):
+        ranges = []
+        for entry in cfg.CONF.ml2_type_vxlan.vni_ranges:
+            entry = entry.strip()
+            try:
+                tun_min, tun_max = entry.split(':')
+                tun_min = tun_min.strip()
+                tun_max = tun_max.strip()
+                ranges.append(range(int(tun_min), int(tun_max) + 1))
```
swagner-de, line 114: We are fine with the fact that only increasing ranges work and decreasing ranges would be not considered?

swagner-de (approved):

networking_ccloud/ml2/agent/common/agent.py
sebageek: Yes, I think so. The tunnel range config is verified by OpenStack [in the ml2 type driver fortunnel](https://github.com/sapcc/neutron/blob/5f8a922d2e7625afb268f834cf7e46f82ff4bf85/neutron/plugins/ml2/drivers/type_tunnel.py#L133-L146) used by vxlan. This calls [verify_tunnel_range()](https://github.com/sapcc/neutron-lib/blob/4d9e219f0135acec031238cbecfba3ead84458e5/neutron_lib/plugins/utils.py#L115-L119) in which we check that the order is ascending.

PR #151, opened 2026-06-08
Preparations for NXOS agent

sebageek:
These are changes on the base code that were all written in the scope of preparing for NXOS. I rewrote bigger parts of the config generator, adjusted the config checker for the k8s-style objects that we generally need to check, implemented the BGW feature we already have on ACI side and modernized some parts. Please see the individual commits for details.


networking_ccloud/tools/config_check.py
```diff
@@ -13,51 +13,82 @@
...
+                    "that is also done by the driver on startup."
     )
-    parser.add_argument("-c", "--config-file")
-    parser.add_argument("-y", "--yaml-file")
-    parser.add_argument("--credentials-file")
+    parser.add_argument("-d", "--driver-config", required=True, help="Path to driver config yaml")
+    parser.add_argument("--ud", "--unwrap-driver-config",
+                        help="Unwrap config dict keys (separated with /), useful when reading config from a ConfigMap")
+    parser.add_argument("--credentials-config", help="Path to driver credentials yaml")
+    parser.add_argument("--uc", "--unwrap-credentials", help="Unwrap credentials dict keys (separated with /)")
```
toanju, line 36: aren't --ud and --uc more like filters for a certain path of the respective dict? At least naming wise this could be hinted here

networking_ccloud/tests/unit/services/trunk/test_driver.py
```diff
@@ -0,0 +1,304 @@
+# Copyright 2022 SAP SE
```
toanju:
```suggestion
# Copyright 2026 SAP SE
```

networking_ccloud/tests/unit/services/trunk/test_driver.py
```diff
@@ -0,0 +1,304 @@
...
+                trunk = {'port_id': trunk_port2['port']['id'],
+                         'project_id': 'test_tenant',
+                         'sub_ports': []}
+                trunk2 = self.trunk_plugin.create_trunk(ctx, {'trunk': trunk})
+                self.assertEqual({trunk1['id'], trunk2['id']},
+                                 set(self.mech_driver.fabric_plugin.get_trunks_with_binding_host(ctx,
+                                                                                                 "node101-seagull")))
+
+    def test_create_trunk_for_hostgroup_with_subport_that_collides_with_metagroup(self):
+        with self.port(device_id='aaa-bbb-ccc') as trunk_port, self.port() as subport:
```
sven-rosenzweig, line 158: Can we override the allowed vlan range here? I think it would be easier to spot why the test fails. 
```
cfg.CONF.set_override('network_vlan_ranges', ['seagull:23:42'],
                              group='ml2_type_vlan')
```

networking_ccloud/tests/unit/services/trunk/test_driver.py
```diff
@@ -0,0 +1,304 @@
...
+            with self.port() as trunk_port2:
+                self._update_vif_type(ctx, trunk_port2, host='node001-seagull')
+                trunk = {'port_id': trunk_port2['port']['id'],
+                         'project_id': 'test_tenant',
+                         'sub_ports': []}
+                self.assertRaisesRegex(nl_cb_exc.CallbackFailure,
+                                       f"Host node001-seagull already has trunk {first_trunk_id} connected to it",
+                                       self.trunk_plugin.create_trunk, ctx, {'trunk': trunk})
+
+    def test_create_trunk_for_hostgroup_which_already_has_a_trunk_but_this_time_it_is_allowed(self):
```
sven-rosenzweig, line 140: For me it was difficult to spot the difference between 
``test_create_trunk_for_hostgroup_which_already_has_a_trunk`` and ``def test_create_trunk_for_hostgroup_which_already_has_a_trunk_but_this_time_it_is_allowed``.  
Somehow indicating ``allow_multiple_trunk_ports=True`` is set for the host ``node101-seagull``, would have helped me. Also ``node001-seagull`` and ``node101-seagull`` is very close from a naming perspective.

networking_ccloud/tools/config_check.py
sebageek: They are not really filters, it's a dict-key path that the checker needs to unpack before the config can be checked. In our config repo we keep the config as a yaml that looks like the ConfigMap, meaning it is wrapped in `cc_fabric --> driver_config --> $actual_config`. With `--ud cc_fabric/driver_config` we would first unwrap this and then check the inner config. Does that make sense?

networking_ccloud/tests/unit/services/trunk/test_driver.py
sebageek: The tests were - I think - written in 2023 and back then I just forgot to add them to git. I can bump the year, though :)

networking_ccloud/tests/unit/services/trunk/test_driver.py
sebageek: We need to set the vlan range for all tests, else all tests would fail. I can add a comment or add an extra override, though the override might need to be already available on driver load in `setUp()`.

networking_ccloud/tests/unit/services/trunk/test_driver.py
sebageek: I renamed the `node101-seagull` to `node-multitrunk-seagull` to make the difference clearer. Does that work?

sven-rosenzweig (approved):

toanju (approved):

PR #154, opened 2026-07-02
Fall back to region ASN for stretched RT

sebageek:
On NXOS 0 is an invalid admin RT value. We might want to go with an explicit value here in the future (like 10), but for now we just fall back to the global ASN from the driver config.


occamshatchet (approved):
Looks good!

toanju (approved):

PR #155, opened 2026-08-03
Improve IPv6 compatibility

sebageek:
In the communication between the ml2 driver and switch agent we did not split up between ipv4 and ipv6 - all ips went into the same lists/attributes and it was up to the agent to split this. For BGP networks and aggregates this works fine, but for the interface primary ip the agent would need to pick a second primary ip from the address family not being the primary ip. To make this clearer we split up primary_ip and secondary_ips into a v4/v6 version and split the IPs in the ml2 driver part.

When sorting ip addresses from both address families we now compare their int value for a stable sorting, as we cannot compare v4 and v6 representations with each other. For gateways we sort the list of each gateway individually to get a stable list.

For general v6 compatibility we now also filter out v6 in the EOS agent for now, until we either require that feature or get a testbed for working on this.

### DIFF
--- a/networking_ccloud/tests/unit/ml2/agent/common/test_messages.py
+++ b/networking_ccloud/tests/unit/ml2/agent/common/test_messages.py
@@ -11,6 +11,7 @@
 #    WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied. See the
 #    License for the specific language governing permissions and limitations
 #    under the License.
+import pydantic
 
 from networking_ccloud.common.config import config_driver
 from networking_ccloud.ml2.agent.common import messages as agent_msg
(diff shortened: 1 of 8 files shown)


networking_ccloud/tests/unit/ml2/agent/common/test_messages.py
```diff
@@ -71,6 +72,27 @@ def test_rt_validation(self):
         self.assertEqual("65130.23:1234", agent_msg.validate_route_target("4268359703:1234"))
         self.assertEqual("123:123", agent_msg.validate_route_target("123:123"))
 
+    def test_vlan_iface_af_check_working(self):
+        agent_msg.VlanIface(
+            vlan=1000,
+            primary_ip_v4="10.100.1.0/24", secondary_ips_v4=["10.100.2.0/24", "10.100.3.0/24"],
+            primary_ip_v6="fe80::1/64", secondary_ips_v6=["fe81::1/64", "fe82::1/64"],
+        )
+
```
mutax: it might be wise to also check if the arguments are actually being used and present in the iface

mutax (dismissed):

mutax (approved):
lgtm

